PR #21, opened 2021-11-25
Fix omnisharp warnings

zhiburt:
changing null to "" may only hide an inner issue.
But at least it may be seen here.

There are 2 different implementations of `IRedisConnection`
both in `Redis.OM` namespace.
I guess this must be addressed as well but I am not sure how.

https://github.com/redis/redis-om-dotnet/blob/main/src/Redis.OM.POC/RedisConnectionExtensions.cs#L10

Can be closed if considered irrelevant.

### DIFF
--- a/src/Redis.OM.POC/RespHelper.cs
+++ b/src/Redis.OM.POC/RespHelper.cs
@@ -89,7 +89,7 @@ namespace Redis.OM
         public static async Task<RedisReply> ReadBulkStringAsync(Socket socket, string currentString)
         {
             if (currentString == "-1")
-                return null;
+                return "";
             var size = int.Parse(currentString);
             var buffer = new ArraySegment<byte>(new byte[size + 2]);
             //var buffer = new byte[size + 2];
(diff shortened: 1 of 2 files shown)


src/Redis.OM.POC/RespHelper.cs
```diff
@@ -89,7 +89,7 @@ public static async Task<int> GetPositionOfNextCRLFAsync(Socket socket)
         public static async Task<RedisReply> ReadBulkStringAsync(Socket socket, string currentString)
         {
             if (currentString == "-1")
-                return null;
+                return "";
```
slorello89, line 92: The redis protocol specifies that a -1 maps to null - so IMO this is correct - only thing I can think to do would be to update the return type from `Task<RedisReply>` -> `Task<RedisReply?>`

slorello89 (changes requested):
Comment about -1 -> null mapping in the reply

src/Redis.OM.POC/RespHelper.cs
zhiburt: Seems good option.

zhiburt: But actually it brings changes in public API.
Specifically this method. must be also made `Nullable`.
Which involves a list a changes is this OK then?

```c#
        Task<RedisReply> ExecuteAsync(string command, params string[] args);
```

PR #22, opened 2021-11-25
Refactoring RedisCollectionStateManager

zhiburt:
Overall I am not as satisfied with changes
but creation of one more Interface seems to much I am even don't have a name for it :sweat_smile: 

So here it is.

Can be closed if considered irrelevant.


slorello89:
Yeah I think let's let this one alone :) I'm going to go ahead and close this - thanks for opening it though!

PR #24, opened 2021-11-29
Docs adjustments

BenShapira:
1. Removed the Index attribute from the `Email` property, that is to show that not all properties require to be indexed.

2. Removed the `Sortable` attribute from the `LastName` property, that is to show that not all Index attribute requires such property.

3. Adjusted the creation of the Redis index to directly access `IRedisConnection` through the `RedisConnectionProvider`.
That is to avoid confusion of not knowing how to access `IRedisConnection`
The current documentation is missing the `connection` variable creation, such as - 
As working directly on the `IRedisConnection` shouldn't be needed in such a wrapper I simply skipped that variable and used `Provider.Connection` directly.

4. Added an example of inserting a new document, that is to show that an insert doesn't need to be made on `IRedisConnection` via some of the Set commands, as an abstraction is available using `IRedisCollection` directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,8 +86,8 @@ With Redis OM, you can model your data and declare indexes with minimal code. Fo
 public class Customer
 {
    [Indexed(Sortable = true)] public string FirstName { get; set; }
-   [Indexed(Sortable = true)] public string LastName { get; set; }
-   [Indexed] public string Email { get; set; }
+   [Indexed] public string LastName { get; set; }
+   public string Email { get; set; }
    [Indexed(Sortable = true)] public int Age { get; set; }
 }
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -86,8 +86,8 @@ With Redis OM, you can model your data and declare indexes with minimal code. Fo
 public class Customer
 {
    [Indexed(Sortable = true)] public string FirstName { get; set; }
-   [Indexed(Sortable = true)] public string LastName { get; set; }
-   [Indexed] public string Email { get; set; }
+   [Indexed] public string LastName { get; set; }
```
slorello89, line 89: Can you make `LastName` Aggregatable? the reason I left it sortable in this example is because a field needs to either be marked sortable or aggregatable to be used in aggregations (which it is further down in the readme)

slorello89 (changes requested):
Just one minor comment.

slorello89 (approved):
LGTM 👍 thanks for contributing!

PR #29, opened 2021-12-03
rename for better dotnet  naming style

witskeeper:
use  `xxxxExtensions` for  static extension class 
put   `ServiceCollectionExtensions` in namespace `Microsoft.Extensions.DependencyInjection`  easy use in `Startup.cs`
use   `xxxx.AspNetCore` for project which dependency `AspNetCore`


src/Redis.OM.AspNetCore/ServiceCollectionExtensions.cs
```diff
@@ -1,11 +1,10 @@
 ﻿using Microsoft.AspNetCore.Builder;
 using Microsoft.Extensions.Configuration;
-using Microsoft.Extensions.DependencyInjection;
 using Redis.OM;
 
-namespace Redis.Developer.AspDotnetCore
+namespace Microsoft.Extensions.DependencyInjection
```
WeihanLi, line 5: Suggest to keep using the library namespace to avoid conflicts with other libraries

witskeeper: Both options are ok.

witskeeper: already use namespace  `Redis.OM.AspNetCore`

PR #34, opened 2021-12-08
Refactoring RedisObjectHandler.SetId, Added Integer to the supported RedisIdFieldAttribute

BenShapira:
1. Adjusted to an easier-to-maintain types check.
2. Added Int to the supported RedisIdFieldAttribute fields, Integer sounds like a commonly used property for indexing, why would Integers not be accepted?
3. Removed non-working default check on an object type, still checking for null value although it covers only strings.
4. Removed idProperty.SetValue(obj, id) to avoid exceptions when the property is not of type string.

### DIFF
--- a/src/Redis.OM/RedisObjectHandler.cs
+++ b/src/Redis.OM/RedisObjectHandler.cs
@@ -101,35 +101,21 @@ namespace Redis.OM
                 throw new MissingMemberException("Missing Document Attribute decoration");
             }
 
-            string id;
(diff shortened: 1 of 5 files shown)


slorello89 (approved):
Pushed a commit to do the following:

* Added a couple of tests and made some 
* tweaks to fix the broken test
* And to make sure that IDs and keys can be created from a valuetype.

LGTM 👍

PR #36, opened 2021-12-13
Added the Insert Object to the Example Application

satish860:
I tried using the POC app and found that the Insert Object was missing. So added the same. 

This is my first pull request to an open-source.  Do let me know if there is any mistake I will correct the same. 

Also, this contribution is really small to make into contributions I can remove my name.


slorello89 (approved):
LGTM 👍 thanks for your contribution @satish860!

PR #39, opened 2021-12-15
Search command Fails When the Index was missing

satish860:
### Context:
I was experimenting with the code and found the search extension method which allows me to work on the low-level API. So I tried the same
``` csharp
SearchResponse<Movies> searchResponse = provider
        .Connection
        .Search<Movies>(new RedisQuery { QueryText = "war" });
    foreach (var item in searchResponse.Documents)
    {
        Console.WriteLine(item.Value.Title);
    }
```
I got an Argument Null exception as IndexName is one of the mandatory parameters. Since my index is automatically created by the library, I took some time to find the index pattern. 

### Solution: 
Since we are already passing the Generics we can use the code

``` csharp 
 query.Index = typeof(T).SerializeIndex().FirstOrDefault() ?? string.Empty;
```
or use the below code
``` csharp
var objAttribute = Attribute.GetCustomAttribute(
        type,
        typeof(DocumentAttribute)) as DocumentAttribute;
    if (objAttribute == null)
    {
        throw new InvalidOperationException($"Type being indexed must be decorated " +
            $"with an RedisObjectDefinitionAttribute, none found on provided type:{type.Name}");
    }

    var args = new List<string>();
    if (string.IsNullOrEmpty(objAttribute.IndexName))
    {
        args.Add($"{type.Name.ToLower()}-idx");
    }
    else
    {
        args.Add(objAttribute.IndexName!);
    }
```

as done in other places. 

**I have not added the Unit test yet. I want to validate the idea to verify if it is done intentionally.**

### DIFF
--- a/src/Redis.OM/RediSearchCommands.cs
+++ b/src/Redis.OM/RediSearchCommands.cs
@@ -23,6 +23,11 @@ namespace Redis.OM
         public static SearchResponse<T> Search<T>(this IRedisConnection connection, RedisQuery query)
             where T : notnull
         {
+            if (string.IsNullOrEmpty(query.Index))
+            {
+                query.Index = typeof(T).SerializeIndex().FirstOrDefault() ?? string.Empty;
+            }
+
             var res = connection.Execute("FT.SEARCH", query.SerializeQuery());
             return new SearchResponse<T>(res);
         }
@@ -37,6 +42,11 @@ namespace Redis.OM
         public static async Task<SearchResponse<T>> SearchAsync<T>(this IRedisConnection connection, RedisQuery query)
             where T : notnull
         {
+            if (string.IsNullOrEmpty(query.Index))
+            {
+                query.Index = typeof(T).SerializeIndex().FirstOrDefault() ?? string.Empty;
+            }
+
             var res = await connection.ExecuteAsync("FT.SEARCH", query.SerializeQuery());
             return new SearchResponse<T>(res);
         }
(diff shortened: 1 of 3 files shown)


src/Redis.OM/RediSearchCommands.cs
```diff
@@ -23,6 +23,11 @@ public static class RediSearchCommands
         public static SearchResponse<T> Search<T>(this IRedisConnection connection, RedisQuery query)
             where T : notnull
         {
+            if (string.IsNullOrEmpty(query.Index))
+            {
+                query.Index = typeof(T).SerializeIndex().FirstOrDefault() ?? string.Empty;
```
slorello89, line 28: That would actually give the Search command an extra output (a morphed Query) I think probably the better way to handle this is to make any constructors for `RedisQuery` take the index name. Technically that's a breaking change but it's very minor and we are still very much still pre-1.0 so I'm fine with pushing it in a pre-1.0 Patch release as an obvious bug-fix.

slorello89 (changes requested):
Force `RedisQuery` to take index name as a constructor parameter.

src/Redis.OM/RediSearchCommands.cs
satish860: Thanks, @slorello89. That's a good suggestion. changes to the code done.

This can help move the error from runtime to compile-time but wondering about the usability.  

To Explain usability better, I decorated the class with the ` Document ` attribute and the index name is autogenerated by the library now I still need to find the index name to provide in the constructor.  Can we do something about that.

slorello89 (approved):
LGTM 👍

PR #40, opened 2021-12-20
Adding Redis Index Info Command 

satish860:
Changes are done to add Get the Redis Info. 

1. Added FT.INFO command and converts to Dictionary of strings
2. Added an async version of the same.
3. Added a Generic method to support the same.


src/Redis.OM/RediSearchCommands.cs
```diff
@@ -130,6 +132,79 @@ public static bool DropIndexAndAssociatedRecords(this IRedisConnection connectio
...
+            var indexName = typeof(T).SerializeIndex().First();
+            return await connection.GetIndexInfoasync(indexName);
+        }
+
+        /// <summary>
+        /// Converts the Redis Reply List to Dictionary.
+        /// </summary>
+        /// <param name="res">Response from the command.</param>
+        /// <returns>Dictionary of Key and Value.</returns>
+        internal static Dictionary<string, string> ToDictionary(RedisReply[] res)
```
slorello89, line 195: Rather than returning a dictionary, we should return an `IndexInfo` object which someone can interrogate the fields of directly, I would say the parsing from a `RedisReply` should be handled directly in there.

src/Redis.OM/RediSearchCommands.cs
```diff
@@ -130,6 +132,79 @@ public static bool DropIndexAndAssociatedRecords(this IRedisConnection connectio
...
+            return connection.GetIndexInfo(indexName);
+        }
+
+        /// <summary>
+        /// Get The Index based on Type.
+        /// </summary>
+        /// <param name="connection">Connection Object.</param>
+        /// <typeparam name="T">Type of object from which the index name is derived.</typeparam>
+        /// <returns>Dictionary of Info.</returns>
+        public static async Task<Dictionary<string, string>> GetIndexInfoasync<T>(this IRedisConnection connection)
```
slorello89, line 184: Response type should be `IndexInfo` rather than a dictionary

src/Redis.OM/RediSearchCommands.cs
```diff
@@ -130,6 +132,79 @@ public static bool DropIndexAndAssociatedRecords(this IRedisConnection connectio
...
+            return info;
+        }
+
+        /// <summary>
+        /// Get The Index based on Type.
+        /// </summary>
+        /// <param name="connection">Connection Object.</param>
+        /// <typeparam name="T">Type of object from which the index name is derived.</typeparam>
+        /// <returns>Dictionary of Info.</returns>
+        public static Dictionary<string, string> GetIndexInfo<T>(this IRedisConnection connection)
```
slorello89, line 172: Response Type should be `IndexInfo` rather than a Dictionary

src/Redis.OM/RediSearchCommands.cs
```diff
@@ -130,6 +132,79 @@ public static bool DropIndexAndAssociatedRecords(this IRedisConnection connectio
...
+            return info;
+        }
+
+        /// <summary>
+        /// Gets the Index with the Index Name Pattern.
+        /// </summary>
+        /// <param name="connection">Connection.</param>
+        /// <param name="indexName">Name of the Index to be queries.</param>
+        /// <returns>Information about the Info.</returns>
+        public static async Task<Dictionary<string, string>> GetIndexInfoasync(
```
slorello89, line 156: Response Type should be `IndexInfo` rather than a Dictionary

src/Redis.OM/RediSearchCommands.cs
```diff
@@ -130,6 +132,79 @@ public static bool DropIndexAndAssociatedRecords(this IRedisConnection connectio
             }
         }
 
+        /// <summary>
+        /// Gets the Index with the Index Name Pattern.
+        /// </summary>
+        /// <param name="connection">Connection.</param>
+        /// <param name="indexName">Name of the Index to be queries.</param>
+        /// <returns>Information about the Info.</returns>
+        public static Dictionary<string, string> GetIndexInfo(this IRedisConnection connection, string indexName)
```
slorello89, line 141: Response Type should be `IndexInfo` rather than a Dictionary

src/Redis.OM/RediSearchCommands.cs
```diff
@@ -1,10 +1,12 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using Redis.OM.Contracts;
 using Redis.OM.Modeling;
 using Redis.OM.Searching;
 using Redis.OM.Searching.Query;
+using StackExchange.Redis;
```
slorello89, line 9: Shouldn't need SE.R imported I wouldn't think?

slorello89 (changes requested):
Hi @satish860 - this is a really solid contribution, my big overarching comment is that I'd rather not have an API reply with a Dictionary when there should be a strongly-typed alternative. The reply from `FT.INFO` should have a relatively defined response that we can map to. See: https://oss.redis.com/redisearch/Commands/#ftinfo

src/Redis.OM/RediSearchCommands.cs
satish860: changed it .

src/Redis.OM/RediSearchCommands.cs
satish860: Changed it

src/Redis.OM/RediSearchCommands.cs
satish860: Changed it.

src/Redis.OM/RediSearchCommands.cs
satish860: Removed the import

src/Redis.OM/RediSearchCommands.cs
satish860: Added Redis Info Object and used the Index to take the property out from Redis Reply. 

I am wondering If I convert it to the dictionary and take it by the Key name so we can assign it to Redis Info.

satish860:
@slorello89  Created a Redis info object with a single property to get an idea. There is one concern is the use of the RedisReply Array index rather than the proper key name for mapping.

So my idea is to convert the info command response to the dictionary of Key and RedisReply first and then move to the Redis Info object. Let me know your view.

satish860:
Declining the Pull request as there is some conflict and not able to resolve the same.

PR #42, opened 2022-01-03
Added the List Async Command.

satish860:
Added the code to get the List of Index. 
* The main idea is to help in case of migration to first check the index by name before we drop or create.

### DIFF
--- a/src/Redis.OM/RediSearchCommands.cs
+++ b/src/Redis.OM/RediSearchCommands.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


slorello89:
Hi @satish860 - I don't believe that `_List` makes a lot of sense to add as a command - in redis parlance the `_` preceding the command name indicates that the command is temporary, I'm not sure what exactly that means but I'd hate to have a newer version of RediSearch effectively break a command of the library because they remove that command in favor a `Scan` command. See the note at the bottom of the `FT._LIST` section of the [docs](https://oss.redis.com/redisearch/Commands/#ft_list). Given that, I'm going to close this PR. Thank you for submitting it though!

PR #43, opened 2022-01-11
Fix StackExchange Connection String formatting

sha3sha3:


slorello89 (changes requested):
See comment

slorello89:
Think this is fixed by #173 so going to close it.

PR #47, opened 2022-01-23
Fixed GeoLoc wrong parsing

ecortese:
Made parsing of longitude and latitude values process culture independent.

Before this fix GeoLocTests.TestParsingFromFormattedHash test fails when process culture define use of comma char as decimal separator for numbers or any other char different from the dot char.

### DIFF
--- a/README.md
+++ b/README.md
@@ -201,6 +201,7 @@ We'd love your contributions! If you want to contribute please read our [Contrib
 * @BenShapira
 * @satish860
 * @dracco1993
+* @ecortese
 
 <!-- Logo -->
 [Logo]: images/logo.svg
--- a/src/Redis.OM/Modeling/GeoLocJsonConverter.cs
+++ b/src/Redis.OM/Modeling/GeoLocJsonConverter.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Globalization;
 using System.Text.Json;
 using System.Text.Json.Serialization;
 
@@ -25,7 +26,8 @@ namespace Redis.OM.Modeling
                 var split = str.Split(',');
                 if (split.Length == 2)
                 {
-                    if (double.TryParse(split[0], out var lon) && double.TryParse(split[1], out var lat))
+                    if (double.TryParse(split[0], NumberStyles.Number, CultureInfo.InvariantCulture, out var lon) &&
+                        double.TryParse(split[1], NumberStyles.Number, CultureInfo.InvariantCulture, out var lat))
                     {
                         return new GeoLoc(lon, lat);
                     }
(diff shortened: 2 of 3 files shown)


ecortese:
Please, stand-by to accept the pull request (if you find it useful) because I'm working of a more general solution to the same problem because it affect also other metods and related tests.
Thank you!

slorello89 (changes requested):
Hey @ecortese thanks for opening this, couple of questions

1: Since the first operation the parser does before parsing the lon/lat out of redis is a string split on a `,`  - wouldn't that throw off the rest of the parsing as you would be hitting 4 strings rather than 4 if you are using a comma as a decimal separator?

2: RediSearch requires that Geo fields be the longitude and latitude split by a `,` - It also appears that it does not allow a comma as a decimal separator, it appears to result in indexing failures. Therefore, if you are reading a geoloc out of Redis (which was the primary intent behind this) you should not encounter this issue. So I guess the primary use of this is to allow GUI input and ship it to your back-end and parse it?

3: Can you add a unit test to validate this change?

4: Please add your name to the Contributors section of the readme

ecortese:
Hi slorello89,

about your first and second question the issue on parsing floating values when the process culture use a non dot char as decimal separator (eg. it-IT use the comma char) does not refer to "input" or "output" values for RediSearch methods (and consequently Redis) but it impacts when floating values are returned as string (as single value or a couple or more values separated by a comma) and these are parsed as double (or any other non-integer numeric type) to instantiate an object (eg. the GeoLoc object). Input and output sintax must not change, as you pointed out.

To better explain: parsing of string "1.5" as double in a process running in an enviroment with it-IT culture return 15

About your 3th question, I didn't provided a test because there is already a test (GeoLocTests.TestParsingFromFormattedHash) that fails under this circumstance (the enviroment's culture in which the process runs...), my PR fix it. Anyway I think that I could write a test that set a different process culture and check that results are not affected.

Lastly, I will add my name with pleasure under the Contributors section when there will be at least one my PR accepted :)

slorello89:
Hi @ecortese - thanks for the detailed explanation :) that makes total sense to me and now I see where you were coming from. I'll maintain the last two asks though:

1. Add a test to demonstrate what has been fixed (basically it should fail with v0.1.4 on my machine but pass when your update is applied) 
2. Add your name to the README so you can be marked as a contributor.

Also I suppose the generalized solution you mentioned above, let me know when that is ready and I will review everything.

Thanks for your contribution!

slorello89 (approved):
LGTM 👍 thanks for your contribution @ecortese !

ecortese:
Thanks @slorello89. I'm still working to a general solution to fix some tests failures due to the same problem when runs under a culture using not dot char decimal separator for numbers. As soon as it will be ready I'll open a new PR.

PR #52, opened 2022-01-28
Enforced culture invariant compliance

ecortese:
This PR fixes all tests that fails when executed in an environment using a culture where the decimal number separator character isn't a dot char (e.g. it-IT, fr-FR, etc.).
For any failing test I've added a namesake one with the suffix "TestingInvariantCultureCompliance" where test runs on different process culture.

### DIFF
--- a/src/Redis.OM/Common/ExpressionParserUtilities.cs
+++ b/src/Redis.OM/Common/ExpressionParserUtilities.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
 using System.Linq.Expressions;
 using System.Reflection;
@@ -16,6 +17,9 @@ namespace Redis.OM.Common
     /// </summary>
     internal static class ExpressionParserUtilities
     {
+        private const string SYSTEMDOUBLETYPE = "System.Double";
+        private const string SYSTEMSTRINGTYPE = "System.String";
+
         /// <summary>
         /// Get's the operand string.
         /// </summary>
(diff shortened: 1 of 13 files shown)


src/Redis.OM/Common/ExpressionParserUtilities.cs
```diff
@@ -401,5 +402,33 @@ private static string TranslateContainsStandardQuerySyntax(MethodCallExpression
             var literal = GetOperandStringForQueryArgs(exp.Arguments[0]);
             return $"{memberName}:{literal}";
         }
+
+        private static string ValueToString(object value)
+        {
+            return value.GetType().FullName switch
```
slorello89, line 408: For all these here, probably do not want to do string comparisons with the fullname of the class, rather want to check against typeof(double), and so forth.

slorello89 (changes requested):
@ecortese , thanks for submitting this, sorry it's taken a bit to get back to you, looks good, only comments I have are around how you are switching based on the full name of the class rather than it's type. Also probably want to account for nullable types as well in those same spots.

src/Redis.OM/Common/ExpressionParserUtilities.cs
ecortese: Hi @slorello89, yes I can see how much better is to use type comparison instead of name. Sincerelly I don't remember why I choiced to do so. I'll work on your change request.

slorello89 (approved):
@ecortese - apologies for taking so long to get back to you on this, I appreciate you bearing with me :) I pushed an update to manage the merge for the drift that's happened in the 8 months since you opened this - it looks good to go for me now - much appreciated.

ecortese:
@slorello89 no problem, have a nice day!

Issue #54, opened 2022-01-30
Pass existing IConnectionMultiplexer object into RedisConnectionProvider constructor


yorfmeister69:
How about we pass an existing IConnectionMultiplexer connection to the RedisConnectionProvider 's constructor? Most of the times, we already have a IConnectionMultiplexer object already injected to our DI. So injecting a new RedisConnectionProvider to our DI, just creates 2 instances of the IConnectionMultiplexer object since the RedisConnectionProvider internally created another IConnectionMultiplexer object.

yorfmeister69:
Any updates on this? Should I create a PR for this?

slorello89:
It's something I was initially skeptical about (essentially exposing another library's classes through Redis OM) but you have a solid point, there's no way to use a pre-existing Multiplexer in Redis OM .NET at the moment, which is not ideal for StackExchange.Redis. Fortunately, looks like your cause was picked up by #59, so it should be in there pretty soon.

slorello89:
Resolved in v0.1.7, so I'm closing this issue.

PR #59, opened 2022-02-01
Updating Connection Provider

baldutech:
The main goal of this pull request is to give a easy way to create different forms of connection.

 - Creating a new way to instance connection provider and give the control of dependency injection, so the whoever consumes can control the resilience for the scope of connection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,7 @@ We'd love your contributions! If you want to contribute please read our [Contrib
 * @satish860
 * @dracco1993
 * @ecortese
+* @baldutech
 
 <!-- Logo -->
 [Logo]: images/logo.svg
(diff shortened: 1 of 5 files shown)


slorello89 (changes requested):
Hello @baldutech - thanks for submitting this

I was initially skeptical of exposing any of the data types from StackExcahnge.Redis to the users of this library but I think that #54 makes a good point insofar as not wanting to duplicate multiplexers (which would just not be any good).

Just two asks if you have the time:

1: Add a functional test in the UnitTest project so we have an automated proof
2: Roll back the changes in the Console App as they should be superfolous with the addition of aforementioned test.

thanks again for contributing!

slorello89 (approved):
LGTM 👍

PR #60, opened 2022-02-01
Exception handling for index creation/deletion

DanJRWalsh:
Added missing exception handling to `CreateIndex`, `CreateIndexAsync` and `DropIndexAsync`

### DIFF
--- a/src/Redis.OM/RediSearchCommands.cs
+++ b/src/Redis.OM/RediSearchCommands.cs
@@ -49,9 +49,21 @@ namespace Redis.OM
         /// <returns>whether the index was created or not.</returns>
         public static bool CreateIndex(this IRedisConnection connection, Type type)
         {
-            var serializedParams = type.SerializeIndex();
-            connection.Execute("FT.CREATE", serializedParams);
-            return true;
+            try
+            {
+                var serializedParams = type.SerializeIndex();
+                connection.Execute("FT.CREATE", serializedParams);
+                return true;
+            }
+            catch (Exception ex)
+            {
+                if (ex.Message.Contains("Index already exists"))
+                {
+                    return false;
+                }
+
+                throw;
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Redis.OM/RediSearchCommands.cs
```diff
@@ -130,6 +142,31 @@ public static bool DropIndexAndAssociatedRecords(this IRedisConnection connectio
             }
         }
 
+        /// <summary>
+        /// Deletes an index. And drops associated records.
+        /// </summary>
+        /// <param name="connection">the connection.</param>
+        /// <param name="type">the type to drop the index for.</param>
+        /// <returns>whether the index was dropped or not.</returns>
+        public static async Task<bool> DropIndexAndAssociatedRecordsAsync(this IRedisConnection connection, Type type)
```
slorello89, line 151: I'm thinking we really don't need the `DropIndexAndAssociatedRecordsAsync` method as this is really something you shouldn't be doing in a situation where sync/async is of major importance 😂.

slorello89 (changes requested):
Thanks for submitting this PR, yeah I'm good with fixing DropIndexAsync as it's going to toss an exception out as if the index already exists, which sort of defeats the purpose of the boolean. 

Please add a functional test to validte this fix, should be fairly trivial to add.

However with the `DropIndexAndAsociatedRecords` I'm skeptical of adding an async version of that method, that method (and the index creation/deletion methods in general) are methods that really shouldn't be run in a situation where sync vs async code is important, so let's pull out the `DropIndexAndAssociatedRecordsAsync` method

src/Redis.OM/RediSearchCommands.cs
DanJRWalsh: @slorello89 That's fair but with that logic should we not remove `DropIndexAsync` then? I was adding it more for consistency than anything.

slorello89: Hi @DanJRWalsh Yeah, I think that's fair please by all means go ahead and add that.

DanJRWalsh:
@slorello89 Let me know if it needs anything else

slorello89 (approved):
LGTM 👍

Issue #114, opened 2022-05-25
GroupBy round time, sum MatchVolume


giangcoi48k:
my model:
```
[Document]
public class IndexSeries
{        
	[RedisIdField, Indexed] public string Id => $"{IndexId}_{TradingDate}";
	[Indexed]public int IndexId {get;set;}
	[Indexed(Sortable = true)]public long TradingDate {get;set;}  
	public decimal OpenIndex { get; set; }
	public decimal CloseIndex { get; set; }
	public decimal HighestIndex { get; set; }
	public decimal LowestIndex { get; set; }
	public long TotalMatchVolume { get; set; } 
	public long MatchValue { get; set; }
	public long MatchVolume { get; set; }    
        ...  
}
```
`TradingDate` is an unix timestamp, round to the second, data received every second.
I want to group data from that index, with a round-up trading date of 1 minute, 5 minutes, 30 minutes..., each group has data by rule:
- `MatchValue `is the sum of `MatchValue` in the group
- `MatchVolume`is the sum of `MatchVolume` in the group
- other data is the latest data in group

```
var aggs = _provider.AggregationSet<IndexSeries>();
var timeStamp = Convert.ToInt64(TimeSpan.FromMinutes(5).TotalMilliseconds); // group by 5 minutes
aggs.Apply(t => (t.RecordShell!.TradingDate + timeStamp - 1) / timeStamp * timeStamp, "TimeGroup")
     .GroupBy(x => x["TimeGroup"])
     ...
```
The above code applied round-up time to the "TimeGroup" group. But now, I don't know how to get the data by the rule that I wanted.
Is there any way to do that? please help!!!

giangcoi48k:
@slorello89 I wonder with that data structure, should I use RedisTimeSeries. Does Redis support the Multivariate time series as above?

slorello89:
RedisTimeseries would be appropriate for this case. You would need to use two separate timeseries to support this and you would need to set up 2 rules, but that solution is purpose built for what you are looking to do.

Issue #113, opened 2022-05-25
Multi-level sorting with Aggregations


diogonborges:
This doesn't look to be documented. So i'm really not sure.

How would I be able to perform multiple level sortby with FT.Aggregate?

Example Model:
```
[Document(IndexName = "employees_idx", StorageType = StorageType.Hash)]
public class Employee
{
    [Searchable(Sortable = true)] public string Name { get; set; }

    [Indexed(Aggregatable = true, Sortable = true)]
    public int Age { get; set; }

    [Indexed(Aggregatable = true)] public double Sales { get; set; }

    [Searchable(Aggregatable = true)] public string Location { get; set; }

    [Indexed(CascadeDepth = 1)] public Department Department { get; set; }
}

public class Department
{
    [Searchable(Aggregatable = true)] public string Name { get; set; }

    [Indexed(Aggregatable = true)] public int Counter { get; set; }
}

```
Example: 
```
FT.AGGREGATE "employees_idx" "*" LOAD * SORTBY 4 @Age ASC @Name ASC
```

This would be roughly equivalent to:
```
FT.SEARCH "employees_idx" "*" SORTBY Age ASC
```

The only way I managed to get something going is by using the Execute method:

```
var reply = provider.Connection.Execute("FT.AGGREGATE", "employees_idx", "*", "LOAD", "*", "SORTBY", "4", "@Age", "ASC", "@Name", "DESC");
```

My main concern is how to do multi-level sorting with this library, particularly without having to use raw commands.
As a side note, from what I'm reading would _have_ to be done with FT.AGGREGATE because of a current limitation of Redis.Search that dropped the support for multi sort on FT.SEARCH. 

Thanks

slorello89:
@diogonborges if I'm being honest I didn't realize we supported mult-level searches in anyway, I've opened a PR in #115 to add them. There is no LOAD predicate support atm, I keep seeing cases though where it would be useful (even though it does add a lot of overhead), so let me poke a bit at it today and see about adding it.

slorello89:
Coming in #122

slorello89:
Closing as #115 and #122 should address all the issues raised here.

PR #115, opened 2022-05-25
Adding multiple-sort support, fixing apply expansion issue.

slorello89:
This should address #113 - adds multiple-sort predicate support e.g.

`collection.OrderBy(x => x.RecordShell.Name).OrderByDescending(x => x.RecordShell.Age)`

should translate to 

`Execute("FT.AGGREGATE","person-idx", "*", "SORTBY", "4", "@Name", "ASC", "@Age", "DESC", "WITHCURSOR", "COUNT", "10000")`

Also addresses an apply expression expansion issue I stumbled over while investigating #114

### DIFF
--- a/src/Redis.OM/Aggregation/AggregationPredicates/AggregateSortBy.cs
+++ b/src/Redis.OM/Aggregation/AggregationPredicates/AggregateSortBy.cs
@@ -36,11 +36,16 @@ namespace Redis.OM.Aggregation.AggregationPredicates
         /// </summary>
         public int? Max { get; set; }
 
+        /// <summary>
+        /// gets the number of arguments of this predicate.
+        /// </summary>
+        internal int NumArgs => Max.HasValue ? 4 : 2;
+
         /// <inheritdoc/>
         public IEnumerable<string> Serialize()
         {
             var numArgs = Max.HasValue ? 4 : 2;
-            var ret = new List<string> { "SORTBY", numArgs.ToString(), $"@{Property}", Direction == SortDirection.Ascending ? "ASC" : "DESC" };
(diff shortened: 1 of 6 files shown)


diogonborges:
Honest question: why would "WITHCURSOR" be added by default?
Wouldn't translating into the the minimum possible command be ideal? as in less is more?

src/Redis.OM/Aggregation/AggregationPredicates/AggregateSortBy.cs
```diff
@@ -36,11 +36,16 @@ public AggregateSortBy(string property, SortDirection direction, int? max = null
...
 
+        /// <summary>
+        /// gets the number of arguments of this predicate.
+        /// </summary>
+        internal int NumArgs => Max.HasValue ? 4 : 2;
+
         /// <inheritdoc/>
         public IEnumerable<string> Serialize()
         {
             var numArgs = Max.HasValue ? 4 : 2;
```
shacharPash, line 47: I think this line is unnecessary because you added the line `internal int NumArgs => Max.HasValue ? 4 : 2;`, and I didn't see that you use numArgs in the code. isn't it?

test/Redis.OM.Unit.Tests/RediSearchTests/ApplyFunctionTests.cs
```diff
@@ -1149,5 +1149,28 @@ public void TestGeoDistanceNested()
...
+
+            _mock.Verify(x => x.Execute(
+                    "FT.AGGREGATE",
+                    "person-idx",
+                    "*",
+                    "APPLY",
+                    expectedPredicate,
+                    "AS",
+                    "res"));
+            Assert.Equal("Blah", res[0]["FakeResult"]);
```
shacharPash, line 1172: You can delete this line if there is the Verify, no?

shacharPash (approved):

PR #132, opened 2022-06-14
Updating Redis Dependency 2.5.61 + Cosmetic Changes for Test Names

shacharPash:


slorello89 (approved):
LGTM 👍 , I think the actual dependency update was made in a previous PR already, In general, I don't think it's worth keeping up along with the mainline StackExchange client unless there are some meaningful goodies in there for us, The last client's we've update to however did have meaningful improvement, bug fixes that can be nicely exploited by Redis OM .NET. (e.g. I think it was somewhere in 2.5 where that weird SELECT issue was addressed).

PR #133, opened 2022-06-15
Fix Time Zone Problem when Read&Write DateTime to Json.

### LINKED ISSUE #69: String with double quote
If a string property of the Model contains a double quote (") the deserialization on file RedisObjectHandler.cs at line 66 fail.
I added a replace in the SendToJson function in file RedisObjectHandler.cs at line 307.

Old:
`ret += $"\"{propertyName}\":\"{hash[propertyName]}\","`

New:
`ret += $"\"{propertyName}\":\"{hash[propertyName].Replace("\"","\\\"")}\","`

I have done that because my string contains a json string.

shacharPash:
fix #69 issue.


slorello89 (approved):
LGTM 👍

PR #136, opened 2022-06-28
Fixing special char query issue for array contains

### LINKED ISSUE #134: Escape @ sign
This will fail with StackExchange.Redis.RedisServerException: 'Syntax error at offset 25 near ca' when using @ sign

```
// Find all customers with the nickname of Jim
customer.Where(x=>x.NickNames.Contains("Jim@earth.com"));'
```

We need to escape string by doing this: 
`customer.Where(x=>x.NickNames.Contains("Jim\\@earth.com"));'`

slorello89:
Need to make sure that when we are doing contains on tag fields that the strings we are checking are properly escaped, fixes #134


shacharPash:
LGTM :+1:

shacharPash (approved):

Issue #165, opened 2022-07-24
Set TTL for the data inserted


MohamadTahir:
Hello, This is more like a feature request,  setting the TTL for keys as an insertion option when inserting new keys.

frostshoxx:
I could use this feature as well. Please let me know if there is anything I can help to expedite the work.

slorello89:
closed by #166

PR #166, opened 2022-07-31
Features/key TTL

shacharPash:
Addresses #165


test/Redis.OM.Unit.Tests/CoreTests.cs
```diff
@@ -45,12 +47,35 @@ public void SimpleGetSetTest()
...
-            var connection = provider.Connection;            
+            var connection = provider.Connection;
+
             connection.Set("x", "value");
             var result = connection.Get("x");
             Assert.Equal("value", result);
         }
 
+        [Fact]
+        public void ExpireTest()
```
slorello89, line 58: Need an async test for this as well

src/Redis.OM/Searching/RedisCollection.cs
```diff
@@ -468,8 +468,13 @@ public async ValueTask SaveAsync()
         }
 
         /// <inheritdoc/>
-        public string Insert(T item)
+        public string Insert(T item, long? expiry = null)
```
slorello89, line 471: Need async counterpart

slorello89, line 471: Also, probably add this as an overload rather than an optional parameter, that way we won't break anyone dependent on the binaries overlaying the dll.

src/Redis.OM/Searching/IRedisCollection.cs
```diff
@@ -38,8 +38,9 @@ public interface IRedisCollection<T> : IOrderedQueryable<T>, IAsyncEnumerable<T>
         /// Inserts an item into redis.
         /// </summary>
         /// <param name="item">an item.</param>
+        /// <param name="expiry">the expiry time of the key in seconds (TTL).</param>
         /// <returns>the key.</returns>
-        string Insert(T item);
+        string Insert(T item, long? expiry = null);
```
slorello89, line 43: Need async version, also change split this out as an overload

src/Redis.OM/RedisCommands.cs
```diff
@@ -156,8 +157,9 @@ public static bool JsonSet(this IRedisConnection connection, string key, string
         /// </summary>
         /// <param name="connection">connection to redis.</param>
         /// <param name="obj">the object to save.</param>
+        /// <param name="expiry">the expiry time of the key in seconds (TTL).</param>
         /// <returns>the key for the object.</returns>
-        public static string Set(this IRedisConnection connection, object obj)
+        public static string Set(this IRedisConnection connection, object obj, long? expiry = null)
```
slorello89, line 162: Needed in async version, also split this out as an overload

src/Redis.OM/RedisCommands.cs
```diff
@@ -177,6 +179,11 @@ public static string Set(this IRedisConnection connection, object obj)
                 connection.JsonSet(id, ".", obj);
             }
 
+            if (expiry != null)
```
slorello89, line 182: This isn't ideal since we are going to be incurring 2 round trips to redis whenever someone tries to execute with an expiration, going to be terrible for performance, these should most likely be executed as part of a batch or transaction rather than as two separate command round trips to redis.

src/Redis.OM/RedisCommands.cs
slorello89, line 162: Probably want two overloads,

1: A true time to live - this might be a timespan rather than a number, could do both so could be three overloads
2: An expire time - this would probably be a date time or another long, either would probably work.

test/Redis.OM.Unit.Tests/Serialization/PersonObject.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using Redis.OM.Modeling;
+
+namespace Redis.OM.Unit.Tests
+{
+    [Document(StorageType = StorageType.Json)]
+    public class Person
```
slorello89, line 7: This shouldn't be necessary, there are Person objects elsewhere in the project, might just need to add an import statement to core tests.

test/Redis.OM.Unit.Tests/CoreTests.cs
```diff
@@ -45,12 +47,35 @@ public void SimpleGetSetTest()
...
+            var provider = new RedisConnectionProvider($"redis://{host}");
+            var connection = provider.Connection;
+
+            Person obj = new Person(){Name = "WithoutExpire"};
+            Person objWithExpire = new Person(){Name = "WithExpire"};
+
+            var objId = connection.Set(obj);
+            var objWithExpireId = connection.Set(objWithExpire, 1);
+
+            Thread.Sleep(2000);
```
slorello89, line 70: probably would want to use `await Task.Delay(2000)` instead of a Thread.Sleep, really not super material since this is just a test, but async over Thread.Sleep is commonly considered best practice.

src/Redis.OM/RedisCommands.cs
```diff
@@ -177,6 +179,11 @@ public static string Set(this IRedisConnection connection, object obj)
...
+            {
+                connection.Execute("EXPIRE", obj.GetKey(), expiry.ToString());
```
slorello89, line 184: `id` is generated above, so there shouldn't be a need to lean on StackExchange's GetKey method.

src/Redis.OM/RedisCommands.cs
```diff
@@ -6,6 +6,7 @@
 using System.Threading.Tasks;
 using Redis.OM.Contracts;
 using Redis.OM.Modeling;
+using StackExchange.Redis;
```
slorello89, line 9: Shouldn't need to import this outside the RedisConnection class.

slorello89 (changes requested):
Few things

1. We need async counterparts of everything we're doing here
2. Running 2 separate Executes is going to be nasty for performance because we're going to be running 2 round trips to Redis, so those need to be run in either a batch or transaction
3. We probably are going to want to use the native data structures of .NET to set expires / expiration times (TimeSpan,DateTime)
4. The methods need to be overloaded rather than have optional parameters added to them. Otherwise, it could be a breaking change to the binaries (not something we need to be super optimized on but its just something that users can run into).

shacharPash:
@slorello89 
Thanks for the comments!
Regarding section 2, what's the alternative?
Because in the JSON.SET and HSET commands I did not see that there is a built-in option of inserting a TTL in the command, which requires doing it in two commands.
I didn't quite understand what you meant when you said "those need to be run in either a batch or transaction".
In the meantime I'm working on the rest of the comments..

src/Redis.OM/RedisCommands.cs
VagyokC4: 2: for completeness, I could see the DateTime & the long overloads.  question, is this long a UnixTime() or UnixTimeMs().  Prefer the later.

slorello89: Nice thing about a timespan is that you can always just convert it to ms - that's actually how they manage it in StackExchange.

test/Redis.OM.Unit.Tests/Serialization/PersonObject.cs
frostshoxx: FYI: There are two places that have class Person defined right now:
- https://github.com/redis/redis-om-dotnet/blob/main/test/Redis.OM.Unit.Tests/RediSearchTests/Person.cs
- https://github.com/redis/redis-om-dotnet/blob/main/test/Redis.OM.Unit.Tests/SearchJsonTests/RedisJsonIndexTests.cs

slorello89 (approved):
LGTM 👍

PR #173, opened 2022-08-06
connection string password bug

rpf3:
* fix a bug with the authenticated connection string generation
* add a .NET editorconfig


rpf3:
@slorello89 I think I have the CI fix in the docker-compose file but I'm not sure the best practice for setting the Redis password in GitHub CI.  Is it fine to commit that in the yaml or is there a better way?

rpf3:
closes #43

.editorconfig
```diff
@@ -0,0 +1,364 @@
+root = true
```
slorello89, line 1: File should be removed

CONTRIBUTING.md
```diff
@@ -16,16 +16,21 @@ We welcome code contributions from the community, if you want to contribute code
 3. Code changes should be relatively small and well described in the pull request.
 4. Code changes should be accompanied by tests that demonstrate coverage and regression for the issue.
 
+### Test Setup
```
slorello89, line 19: Instead of this extra setup, you should just be able to add an extra redis instance in https://github.com/redis/redis-om-dotnet/blob/main/docker/docker-compose.yaml - that requires the password to test the case that requires a password.

slorello89 (changes requested):
I think what you're looking for here is an extra password-protected instance of redis - take a look at the docker folder, that ought to get you what you need.

CONTRIBUTING.md
rpf3: So you mean just use the `PRIVATE_PASSWORD` environment variable in the docker-compose file correct?  How do we then push that down into the CI environment?

rpf3: @slorello89 If I understood the documentation correctly, the idea was to just run a single Redis instance locally for the test suite, expecting that things like the sentinel tests would fail since that is defined in the docker-compose file.  Are you saying you want the private connection tests to follow this pattern?

rpf3: @slorello89 any update on this?

slorello89 (approved):
needed some configuration updates in the docker-compose in order for the tests to work correctly in GH, that's all taken care of now. LGTM 👍 Thanks for the contribution @rpf3

PR #177, opened 2022-08-17
Update Test Project to .NET 6

berviantoleo:
Since .NET is already [out of support](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core#lifecycle), how about upgrading the test project to use .NET 6?

Feel free to give any feedback about these changes.


shacharPash (approved):
LGTM :+1:

PR #185, opened 2022-08-29
Fix #170 to save FindById result to StateManager

### LINKED ISSUE #170: FindById not being added to StateManager
If you `FindById` on the RedisCollection, the returned objects are not being added to the state manager, hence if you try to Save them the state manager does not see them and does not do anything with those objects.

berviantoleo:
Try to fix #170


slorello89 (approved):
LGTM 👍 - thanks for opening this!

Issue #189, opened 2022-08-30
Use Concurrent Dictionary Internally


VagyokC4:
Hi @slorello89,

I came across another interesting result while testing with Redis-OM.  I think I am hitting up against internal structures when trying to do things asynchronously.

Here is my sample code:
```csharp
var categories = RedisDb.RiskCategoryRecord.Where(x => x.Owners.Contains(ERN)).ToListAsync();
var components = RedisDb.RiskComponentRecord.Where(x => x.Owners.Contains(ERN)).ToListAsync();
var factors    = RedisDb.RiskFactorRecord.Where(x => x.Owners.Contains(ERN)).ToListAsync();
var answers    = RedisDb.RiskAnswerRecord.Where(x => x.Owners.Contains(ERN)).ToListAsync();
var lookups    = RedisDb.RiskAnswerRecord.Where(x => x.Owners.Contains(ERN)).ToListAsync();
var overrides  = RedisDb.RiskOverrideRecord.Where(x => x.Owners.Contains(ERN)).ToListAsync();

await Task.WhenAll(categories, components, factors, answers, lookups, overrides).ConfigureAwait(false);
```

Not everytime, but when stress testing I get the following error:
```
An item with the same key has already been added.
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Redis.OM.Searching.RedisCollectionEnumerator`1.ConcatenateRecords()
   at Redis.OM.Searching.RedisCollectionEnumerator`1.GetNextChunkAsync()
   at Redis.OM.Searching.RedisCollectionEnumerator`1.MoveNextAsync()
   at Redis.OM.Searching.RedisCollection`1.ToListAsync()
   at Redis.OM.Searching.RedisCollection`1.ToListAsync()
```

Here I am running it 20 times and you can see sometime I run without error, sometimes I get the error:

https://user-images.githubusercontent.com/253250/187561176-1a40ed92-d21b-435d-9bf0-1bdeda203b45.mp4

So I think this is happening because each call is trying to update an internal dictionary, and at times we get a race condition.  Can we can swap it from a normal dictionary to a Concurrent Dictionary that can be a little more thread safe especially during highly asynchronous use?

slorello89:
[...]

PR #194, opened 2022-09-07
allows users to not save enumerated data in the StateManager

slorello89:
#189 points out an important point, many folks will not want the StateManager to maintain any of their records, whether this is because of the concurrency issues that are pointed out in there, or they just don't want very large collections of data initalized under the hood, giving folks the ability to configure this will be pretty helpful. Adding new path to allow users to decide whether or not to have said data stored.

### DIFF
--- a/src/Redis.OM/RedisConnectionProvider.cs
+++ b/src/Redis.OM/RedisConnectionProvider.cs
@@ -70,5 +70,15 @@ namespace Redis.OM
         /// <returns>A RedisCollection.</returns>
         public IRedisCollection<T> RedisCollection<T>(int chunkSize = 100)
             where T : notnull => new RedisCollection<T>(Connection, chunkSize);
+
+        /// <summary>
+        /// Gets a redis collection.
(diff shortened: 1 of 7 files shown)


shacharPash (approved):

VagyokC4:
@slorello89 Is it possible to have a fluent call that can enable / disable at invocation time?

```csharp
var collection = redisConnectionProvider.RedisCollection<T>();
var itemsWithoutState = await collection.WithoutState().Where(x=>x.IsTrue == true).ToListAsync();
var itemsWithState = await collection.WithState().Where(x=>x.IsTrue == true).ToListAsync();
```

Or something to that effect?

slorello89:
@VagyokC4 - Can you open that suggestion in a new issue? Have to grock on it a bit, but I think where I'd come down is that the extra API seems superfluous seeing as when you have a RedisCollection without state being saved it then becomes ultra-transient and light weight - worth a discussion if nothing else.

Issue #204, opened 2022-09-19
JSON.SET with NX/XX


iozcelik:
Normally Redis has support to `JSON.SET key path value [NX | XX]` command.

I think this have to implement the dotnet library. We use raw command now but it is not best approach. `InsertIfNotExist` or `InsertIfExist` native support will be good. This result could return bool.

`true` if executed correctly, or `false` if the specified NX or XX conditions were not met.

**Conditions**
NX - only set the key if it does not already exist
XX - only set the key if it already exists

shacharPash:
I will work on it soon.

zulander1:
@shacharPash Not sure if it is possible, can we also add: AddorUpdate ? this would be great since it would take care of all the checks

VagyokC4:
> @shacharPash Not sure if it is possible, can we also add: AddorUpdate ? this would be great since it would take care of all the checks

Wouldn't AddOrUpdate() effectively be the same as an Insert()...  or are you think of this more for the statefull side where it would replace or patch on the request?

zulander1:
No, i've created https://github.com/redis/redis-om-dotnet/issues/212

PR #206, opened 2022-09-20
Add nx/xx

shacharPash:
Related to #204


src/Redis.OM/Searching/IRedisCollection.cs
```diff
@@ -69,6 +69,38 @@ public interface IRedisCollection<T> : IOrderedQueryable<T>, IAsyncEnumerable<T>
         /// <returns>the key.</returns>
         Task<string> InsertAsync(T item, TimeSpan timeSpan);
 
+        /// <summary>
+        /// Inserts an item into redis only if it already exists. Valid for Json StorageType only.
+        /// </summary>
+        /// <param name="item">an item.</param>
+        /// <param name="timeSpan">The timespan of the document's (TTL).</param>
+        /// <returns>the key. or null if the key did not exist.</returns>
+        string? InsertIfExist(T item, TimeSpan? timeSpan = null);
```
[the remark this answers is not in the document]
shacharPash: I felt that this way is simpler to implement without breaking too many functions, but you can also use "when".

src/Redis.OM/RedisCommands.cs
```diff
@@ -187,6 +187,45 @@ public static async Task<bool> JsonSetAsync(this IRedisConnection connection, st
...
+        /// Sets a value as JSON in redis.
+        /// </summary>
+        /// <param name="connection">the connection.</param>
+        /// <param name="key">the key for the object.</param>
+        /// <param name="path">the path within the json to set.</param>
+        /// <param name="json">the json.</param>
+        /// <param name="when">XX - set if exist, NX - set if not exist.</param>
+        /// <param name="timeSpan">the the timespan to set for your (TTL).</param>
+        /// <returns>whether the operation succeeded.</returns>
+        public static async Task<bool> JsonSetAsync(this IRedisConnection connection, string key, string path, string json, string when, TimeSpan? timeSpan = null)
```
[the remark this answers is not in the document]
slorello89: yeah, we can add a different enum here, I don't want to expose any data-types from SE that we don't have too.

slorello89:
@chayim - at this point, I've added to much to this PR to be an objective reviewer - can you please take a look?

src/Redis.OM/RedisCommands.cs
```diff
@@ -286,6 +327,48 @@ public static bool JsonSet(this IRedisConnection connection, string key, string
...
+        /// <param name="key">the key for the object.</param>
+        /// <param name="path">the path within the json to set.</param>
+        /// <param name="json">the json.</param>
+        /// <param name="when">XX - set if exist, NX - set if not exist.</param>
+        /// <param name="timeSpan">the the timespan to set for your (TTL).</param>
+        /// <returns>whether the operation succeeded.</returns>
+        public static bool JsonSet(this IRedisConnection connection, string key, string path, string json, WhenKey when, TimeSpan? timeSpan = null)
+        {
+            var argList = new List<string> { timeSpan != null ? ((long)timeSpan.Value.TotalMilliseconds).ToString() : "-1", path, json };
+            switch (when)
```
chayim, line 343: Same question re if/else

src/Redis.OM/RedisCommands.cs
```diff
@@ -315,6 +398,108 @@ public static string Set(this IRedisConnection connection, object obj)
...
+        /// <param name="timespan">The length of time before the key expires.</param>
+        /// <returns>the key for the object, null if nothing was set.</returns>
+        public static string? Set(this IRedisConnection connection, object obj, WhenKey when, TimeSpan? timespan = null)
+        {
+            var id = obj.SetId();
+            var type = obj.GetType();
+
+            if (Attribute.GetCustomAttribute(type, typeof(DocumentAttribute)) is not DocumentAttribute attr || attr.StorageType == StorageType.Hash)
+            {
+                if (when == WhenKey.Always)
```
chayim, line 417: Can these be combined into a single branch? if when == && timespan.HasValue?

Branching multiple portions in the same if statement (rather than sub) is usually more performant.

src/Redis.OM/RedisCommands.cs
```diff
@@ -315,6 +398,108 @@ public static string Set(this IRedisConnection connection, object obj)
...
+                {
+                    if (timespan.HasValue)
+                    {
+                        return connection.Set(obj, timespan.Value);
+                    }
+
+                    return connection.Set(obj);
+                }
+
+                var kvps = obj.BuildHashSet();
```
chayim, line 427: kvps == key value properties?

src/Redis.OM/RedisCommands.cs
```diff
@@ -187,6 +186,48 @@ public static async Task<bool> JsonSetAsync(this IRedisConnection connection, st
...
+        /// Sets a value as JSON in redis.
+        /// </summary>
+        /// <param name="connection">the connection.</param>
+        /// <param name="key">the key for the object.</param>
+        /// <param name="path">the path within the json to set.</param>
+        /// <param name="json">the json.</param>
+        /// <param name="when">XX - set if exist, NX - set if not exist.</param>
+        /// <param name="timeSpan">the the timespan to set for your (TTL).</param>
+        /// <returns>whether the operation succeeded.</returns>
+        public static async Task<bool> JsonSetAsync(this IRedisConnection connection, string key, string path, string json, WhenKey when, TimeSpan? timeSpan = null)
```
chayim, line 199: I'm of two sides here.

On the client side, it's very clear that we have to expose based on name XX vs NX. But here.. I think it's a bad variable name. Ideally, to help users we could just name ifExists or similar. WDYT?

src/Redis.OM/RedisCommands.cs
```diff
@@ -187,6 +186,48 @@ public static async Task<bool> JsonSetAsync(this IRedisConnection connection, st
...
+
+        /// <summary>
+        /// Sets a value as JSON in redis.
+        /// </summary>
+        /// <param name="connection">the connection.</param>
+        /// <param name="key">the key for the object.</param>
+        /// <param name="path">the path within the json to set.</param>
+        /// <param name="obj">the object to serialize to json.</param>
+        /// <param name="when">XX - set if exist, NX - set if not exist.</param>
+        /// <param name="timeSpan">the the timespan to set for your (TTL).</param>
```
chayim, line 223: Why timeSpan and not just seconds or evevn TTL? When I see timeSpan, as a user I assume I need seconds.. but maybe these are milliseconds?

I think the API should be more explict.

src/Redis.OM/RedisCommands.cs
```diff
@@ -286,6 +327,48 @@ public static bool JsonSet(this IRedisConnection connection, string key, string
...
 
+        /// <summary>
+        /// Sets a value as JSON in redis.
+        /// </summary>
+        /// <param name="connection">the connection.</param>
+        /// <param name="key">the key for the object.</param>
+        /// <param name="path">the path within the json to set.</param>
+        /// <param name="json">the json.</param>
+        /// <param name="when">XX - set if exist, NX - set if not exist.</param>
+        /// <param name="timeSpan">the the timespan to set for your (TTL).</param>
```
chayim, line 338: Same for both when + timespan

src/Redis.OM/RedisCommands.cs
```diff
@@ -286,6 +327,48 @@ public static bool JsonSet(this IRedisConnection connection, string key, string
...
+
+        /// <summary>
+        /// Sets a value as JSON in redis.
+        /// </summary>
+        /// <param name="connection">the connection.</param>
+        /// <param name="key">the key for the object.</param>
+        /// <param name="path">the path within the json to set.</param>
+        /// <param name="obj">the object to serialize to json.</param>
+        /// <param name="when">XX - set if exist, NX - set if not exist.</param>
+        /// <param name="timeSpan">the the timespan to set for your (TTL).</param>
```
chayim, line 364: And again

src/Redis.OM/RedisCommands.cs
```diff
@@ -315,6 +398,108 @@ public static string Set(this IRedisConnection connection, object obj)
...
+                var argsList = new List<string>();
+                int? res = null;
+                argsList.Add(timespan != null ? ((long)timespan.Value.TotalMilliseconds).ToString() : "-1");
+                foreach (var kvp in kvps)
+                {
+                    argsList.Add(kvp.Key);
+                    argsList.Add(kvp.Value);
+                }
+
+                if (when == WhenKey.Exists)
```
chayim, line 437: Can this become a straight boolean?

chayim (changes requested):
Logic works, some suggestions on perf and (perceived) API improvements for users.

src/Redis.OM/RedisCommands.cs
slorello89: it can, but you'd be adding an extra comparison:

```cs
if(when == WhenKey.Always && timeSpan.HasValue)
{
// set with timespan
}
else if (when == WhenKey.Always)
{
// set without timespan
}
```

The perf difference again would be negligible here. It's probably faster as is because in the case where `when == WhenKey.Always` you would always be performing two comparisons, versus with the if/else-if where you could be performing as many as 3. And then for the cases where `when != WhenKey.Always` in the current implementation you'd only perform 1 comparison, whereas with the if/else-if you'd be performing 3. Again, this is a super tiny thing, IMO it's better as is.

src/Redis.OM/RedisCommands.cs
slorello89: see above

slorello89 (commented):
Thanks' @chayim - see my responses in this review.

slorello89 (approved):
LGTM 👍

PR #210, opened 2022-09-23
fixing other stateful queries issue

### LINKED ISSUE #205: The collections state is no longer correct. even when saveState = false when using FirstOrDefaultAsync
Hi @slorello89, 

Would you expect to still receive this error even when saveState is set to false?

```
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. 
A concurrent update was performed on this collection and corrupted its state. The collections state is no longer correct.   
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Redis.OM.Modeling.RedisCollectionStateManager.InsertIntoSnapshot(String key, Object value)
   at Redis.OM.Searching.RedisCollection`1.FirstOrDefaultAsync(Expression`1 expression)
```

slorello89:
Fixes #205, adding better exception handling for folks who try enumerating the same redis collection twice. @VagyokC4 - can you check this to make sure this fixes your issue :)

### DIFF
--- a/src/Redis.OM/Aggregation/RedisAggregationSet.cs
+++ b/src/Redis.OM/Aggregation/RedisAggregationSet.cs
@@ -37,7 +37,7 @@ namespace Redis.OM.Aggregation
             }
 
             _chunkSize = chunkSize;
-            Initialize(new RedisQueryProvider(connection, rootAttribute, _chunkSize), null, useCursor);
+            Initialize(new RedisQueryProvider(connection, rootAttribute, _chunkSize, true), null, useCursor);
         }
 
         /// <summary>
--- a/src/Redis.OM/Searching/RedisCollection.cs
+++ b/src/Redis.OM/Searching/RedisCollection.cs
@@ -55,7 +55,7 @@ namespace Redis.OM.Searching
             _connection = connection;
             SaveState = saveState;
             StateManager = new RedisCollectionStateManager(rootAttribute);
-            Initialize(new RedisQueryProvider(connection, StateManager, rootAttribute, ChunkSize), null, null);
+            Initialize(new RedisQueryProvider(connection, StateManager, rootAttribute, ChunkSize, SaveState), null, null);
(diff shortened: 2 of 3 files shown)


VagyokC4:
> Fixes #205, adding better exception handling for folks who try enumerating the same redis collection twice. @VagyokC4 - can you check this to make sure this fixes your issue :)

The commit looks good.  Can you drop 0.2.4 to test?

VagyokC4:
@shacharPash @slorello89 Are we able to push this out to nuget real quick?

VagyokC4:
> @VagyokC4 - can you build it from source to check to make sure it totally solves the problem for you? I'm at a conference this week so a bit OOP :)

@slorello89 Hmmm... so it's working, but it's not.  I'm trying to reduce down to a repeatable test... but the initial bug is gone, as I get my data back, however the connection is now dead and all future queried timeout.  Any ideas?

I'll keep testing and update as I know more.

VagyokC4:
@slorello89 Ok, We can close this issue out, as the fix appears to be working.  I will open a new issue as the issue I am seeing I can reproduce in older versions (0.2.1) so this is a completely different / new issue.

slorello89:
@VagyokC4 - that issue you are describing is an older issue with RediSearch (not redis-om-dotnet), where the client was freezing when you went over 10k search results, that issue SHOULD be resolved already.

slorello89:
@VagyokC4 opened a separate issue and confirmed that his 10K search result issue was fixed in a forthcoming version of redis-stack (only applies to RC-7 I think?)

banker (approved):

slorello89:
Waiting on approvals @rpf3 - ping @shacharPash @chayim

shacharPash (approved):

PR #211, opened 2022-09-23
moving from redismod->redisstack

slorello89:
Moving to redis-stack image from redis mod (needed to make a slight change to unit tests because of default configuration of redis-stack)


K-Jo:
@slorello89 w00t

shacharPash (approved):

Issue #229, opened 2022-10-06
GeoLoc serialized with wrong decimal separator


mariusmuntean:
On my machine `CultureInfo.CurrentCulture.Name` is "en-DE" and `CultureInfo.CurrentCulture.NumberFormat.NumberDecimalSeparator` is ",". 
Because of that, the `GeoLoc` properties on my stored entities use the wrong decimal separator, e.g. "-111,8046654,41,7693747". This breaks the search for these entities completely. Even an `FT.SEARCH` by the ID of such an entity will not return anything.

If I manually replace the decimal separator in the `GeoLoc` of an entity in Redis with "." (with RedisInsight-v2) then the entity shows up in my searches. Also, if I set the default culture and default UI culture to the `InvariantCulture` right before the app starts, the decimal separator will change to "." and everything works as expected.

I think that the `ToString()` method of the `GeoLoc` struct should explicitly use the `InvariantCulture` or any other way to set the decimal separator to ".".

slorello89:
Hi @mariusmuntean

slorello89:
Hi @mariusmuntean - so one of our contributors opened #52 to fix the broader issues of cultural writing of floating points within Redis.OM quite a while ago, I have been sitting on it (mostly because of the drift between main and the PR) this issue SHOULD be resolved with the latest build of Redis OM, can I trouble you to check (if it is I can cut a new release)

mariusmuntean:
I had a look at the PR as well and indeed it looks like it should fix the issue.
How should I go about to check if it fixes the issue on my machine?

slorello89:
try pointing to the .csproj directly from your application and building it from source

mariusmuntean:
Sorry for taking so long, but I wanted to make sure I get this right.
Unfortunately it seems like the problem still persists with `GeoLoc`.

The good news is that the fix is quite easy, just adapt its `ToString()` method to this
```
public override string ToString()
{
  return $"{Longitude.ToString(CultureInfo.InvariantCulture)},{Latitude.ToString(CultureInfo.InvariantCulture)}";
}
```

I could make a PR for this, though it is quite small.

slorello89:
@mariusmuntean - thanks for checking PRs always welcome :) can even tag it as a [hacktoberfest](https://hacktoberfest.com/) PR :)

mariusmuntean:
@slorello89 I gave it a shot in this PR #234

slorello89:
closed via #234

PR #234, opened 2022-10-12
Bugfix for GeoLoc Serialization in Different Cultures

mariusmuntean:
GeoLoc.ToString() doesn't produce a stable output if the culture changes.
This PR fixes that - see #229


slorello89:
Thanks for opening this @mariusmuntean - will push this out :)

PR #236, opened 2022-10-13
Add IRedisCollection Update and Save examples

atrievel:


examples/Redis.OM.UpdatingDocuments/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+coffee.Description = "Pure developer fuel";
+products.Update(coffee);
+
+// update using IRedisCollection.UpdateAsync
+coffee.Price = 2.50;
+await products.UpdateAsync(coffee);
+
+// update using IRedisCollection.Save
+coffee.InStock = true;
+products.Save(coffee);
```
slorello89, line 34: This won't build 😅 - Save doesn't take any arguments, it's meant to allow you to materialize just what you want from the collection, make the updates you'd like, and then save them back to Redis.

examples/Redis.OM.UpdatingDocuments/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+
+// update using IRedisCollection.SaveAsync
+coffee.DateAdded = DateTime.UtcNow.AddDays(-7);
+await products.SaveAsync(coffee);
```
slorello89, line 38: Same with this one.

slorello89 (changes requested):
As it stands right now this example will not build because it's not correctly using the `Save`/`SaveAsync` feature within Redis OM. This should realistically be querying Redis for records it would like to update, update them in the app, and then save them back to Redis with `Save`

slorello89:
Looking pretty good now - couple notes (that I fixed)

1: Tags cannot be marked as sortable in JSON fields (index creation just blows up)
2: Date-times are converted to local time when pushed to redis so your actual/expected were a bit off in the example.

SuzeShardlow:
Thanks for your contribution, @atrievel.  I've given you the `Hacktoberfest 2022` role in our Discord server :)

PR #240, opened 2022-10-19
adding ability to index and query datetimes

slorello89:
Adds ability to index and query DateTimes in Redis with Redis OM, this will be a SOMEWHAT breaking change, because the mode of storage of DateTimes in Redis is changing from strings to numerics, this shouldn't effect anyone only working in Redis OM and interacting with said datetimes because the JSON parser is setup to handle both cases, but all the same it's changing the behavior at storage, so I'm revving to 0.4.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,10 @@ Once the index is created, we can:
 
 Let's see how!
 
+### Indexing DateTimes
+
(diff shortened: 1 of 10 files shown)


bsbodden (approved):
LGTM!

iozcelik:
DateOnly and TimeOnly should included this issue.

Issue #242, opened 2022-10-21
Support dynamic index definition for multi-tenancy (Prefix, IndexName)


houstonhaynes:
This issue relates to [issue 98](https://github.com/redis/redis-om-dotnet/issues/98) so I will try to keep this tight.

We are looking to manage multi-tenancy with discrete indices (or indexes) where there is a customer (company) specific index as well as an "admin" index that allows admins to see all records. So in our case we'll always have N+1 number of indices - one for each customer and the +1 over-arching index that admins can use to see all data. 

ACLs are used to "tune" an individual Redis "user" (in our case a customer/company) to only see the records that have a key with their ID present - and similarly with subscription channels.

The Challenge with OM as it stands now is that index definition is held within the decorator of the type which defines the index. That decorator requires constants, so they are "baked" into the program at compile-time. In a domain where only one index is required this is OK, but in a configuration where multiple tenants are managed with discrete indices and ACLs this is a blocker.

In discussion with my working group a "code gen" approach was suggested - BUT - we arrived at a stopping place with that when we considered the business case of adding a customer (and its respective users) to the identity system where a new ACL and customer-specific index would be required while the application was in-flight. To code-gen multiple types from a base at compile time means that you can only generate types for the companies you can know at compile time. So to add a new customer index would ostensibly require re-compilation.

From this I'd assert that the type decorator is **_not_** the right location for setting index properties such as "PREFIX" and "IndexName". They should be eligible to be assigned at run-time and not limited to a compile-time construct.

VagyokC4:
Hi @houstonhaynes,

We tackled something like this with a simple array property called Owners[] directly on the record.  Then we pushed into this array all owners (ids) who have access to the record.  Then we query based on ids that come from the security context to ensure you only get back records that you are allowed to see.  `x=>x.Owners.Contains("some_id")`

This allowed us to have a single index / property to control who gets access to what etc.  We put an admin id along with tenant id, along with customer id etc. and providing any one of the Owner ids gets back the record.

Maybe something like this works for your use case?

houstonhaynes:
Thanks for this! We considered a similar approach and after considering and experimenting a bit we set it aside as an option on two grounds, 1) security and 2) client burden.

[...]

PR #245, opened 2022-10-26
adding ability to add prefixes to indexes for Searches

slorello89:
Adds overload to to RedisCollection to enable you to create your own prefixes dynamically, this is helpful for cases such as #242 where someone might want to have an index per tenant with it's own set of prefixes.

Basically now you initalize a RedisCollection<T> with a prefix, that prefix then becomes what your Collection uses to:

1. Serialize the index - the index name is generated as {classnametolower}-{prefix}-idx, and of course the prefixe for the index is the specified prefix.
2. At insertion into redis, the objects are given a keyname that is: `{prefix}:{id}`
3. When Querying, the collection uses the new sub-tenants index in the query.

### DIFF
--- a/src/Redis.OM/Aggregation/RedisAggregationSet.cs
+++ b/src/Redis.OM/Aggregation/RedisAggregationSet.cs
@@ -37,7 +37,7 @@ namespace Redis.OM.Aggregation
             }
 
             _chunkSize = chunkSize;
-            Initialize(new RedisQueryProvider(connection, rootAttribute, _chunkSize, true), null, useCursor);
+            Initialize(new RedisQueryProvider(connection, rootAttribute, _chunkSize, true, string.Empty), null, useCursor);
         }
 
         /// <summary>
--- a/src/Redis.OM/Common/ExpressionTranslator.cs
+++ b/src/Redis.OM/Common/ExpressionTranslator.cs
@@ -180,9 +180,10 @@ namespace Redis.OM.Common
         /// <param name="expression">The expression.</param>
         /// <param name="type">The root type.</param>
         /// <param name="mainBooleanExpression">The primary boolean expression to build the filter from.</param>
+        /// <param name="prefix">The index to use when creating the indexName.</param>
         /// <returns>A Redis query.</returns>
         /// <exception cref="InvalidOperationException">Thrown if type is missing indexing.</exception>
-        internal static RedisQuery BuildQueryFromExpression(Expression expression, Type type, Expression? mainBooleanExpression)
+        internal static RedisQuery BuildQueryFromExpression(Expression expression, Type type, Expression? mainBooleanExpression, string prefix)
         {
             var attr = type.GetCustomAttribute<DocumentAttribute>();
             if (attr == null)
@@ -190,7 +191,8 @@ namespace Redis.OM.Common
                 throw new InvalidOperationException("Searches can only be performed on objects decorated with a RedisObjectDefinitionAttribute that specifies a particular index");
             }
 
-            var indexName = string.IsNullOrEmpty(attr.IndexName) ? $"{type.Name.ToLower()}-idx" : attr.IndexName;
+            var prefixAddendum = string.IsNullOrEmpty(prefix) ? string.Empty : $"-{prefix}";
+            var indexName = string.IsNullOrEmpty(attr.IndexName) ? $"{type.Name.ToLower()}{prefixAddendum}-idx" : $"{attr.IndexName}{prefixAddendum}";
             var query = new RedisQuery(indexName!) { QueryText = "*" };
             switch (expression)
             {
--- a/src/Redis.OM/Modeling/RedisIndex.cs
+++ b/src/Redis.OM/Modeling/RedisIndex.cs
@@ -26,9 +26,10 @@ namespace Redis.OM.Modeling
         /// Serialize the Index.
         /// </summary>
         /// <param name="type">The type to be indexed.</param>
+        /// <param name="prefix">The Prefix to use for the index. If this is set, the prefixes property in the DocumentAttribute will be ignored.</param>
         /// <exception cref="InvalidOperationException">Thrown if type provided is not decorated with a RedisObjectDefinitionAttribute.</exception>
         /// <returns>An array of strings (the serialized args for redis).</returns>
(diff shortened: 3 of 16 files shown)


slorello89:
@houstonhaynes @VagyokC4 - what are your thoughts?

houstonhaynes:
I'm confused by what you mean by "prefix".

I took the term to mean "the part of the key that the index applies to" 

Where it looks like you interpret it as "the part of the name that goes before "-idx"

Did I get that right?

What I was originally talking included both the key-part that the index uses to discover/add records. I'll give an example:

Say I have two customers in a Redis instance. One customer has the ID of "A" and the other "B". The keys I generate for the records for each respectively would be "Customer:A:[some-random-GUID]" and "Customer:B:[some-random-GUID]"

Then I would create **three** indices, named "CustomerA-idx" "CustomerB-idx" and "CustomerAdmin-idx". 

CustomerA-idx would look for all keys with the partial key match to "Customer:A:*" 
CustomerB-idx would look for all keys with the partial key match to "Customer:B:*"
CustomerAdmin-idx would look for all keys with the partial key match to "Customer:*"

So in effect - the CustomerA ACL would be set to access only "CustomerA-idx" and so would be tenant limited to Customer "A" records. By extrapolation the CustomerB ACL would limit those users to only see Customer "B" keyed records. And finally - admins would have a CustomerAdmin ACL would would be able to see all keys because they're using an index which matches to the "Customer:*" partial key pattern.

To be more specific - here's a sample of what I would set for the ACL for each company "user":

`ACL SETUSER companya on >coapwd +@all -@dangerous +client +info +echo &company:a: ~company:a:*
ACL SETUSER companyb on >cobpwd +@all -@dangerous +client +info +echo &company:b: ~company:b:*`

And here's what I'd set for the admin role:

`ACL SETUSER admin on >adminpwd +@all -@dangerous +client +info +echo &company: ~company:*`

I hope this clarifies what I'm trying to achieve. This means that both IndexName and Prefix need to be settable at run-time.

slorello89:
@houstonhaynes - So the idea here is that the index's name would be computable given a prefix and the model, the format would now be `classname-Prefix-idx`, so you can easily setup your acls to block them still. What prefix most consequentially is, is a means of determing what your key-prefixes are when they are inserted into redis, the example you provide should work. Given the model

```cs
[Document(StorageType = StorageType.Json, Prefixes = new string[]{"Customer"})]
public class Customer
{
    [RedisIdField] public string Id { get; set; }
    [Indexed] public string FirstName { get; set; }
    [Indexed] public string LastName { get; set; }
}
```

you would then be able to do something like:

```cs
// connect
connection.Execute("FLUSHDB");

var customers = provider.RedisCollection<Customer>();
var customersA = provider.RedisCollection<Customer>(false, 1000, "Customer:A");
var customersB = provider.RedisCollection<Customer>(false, 1000, "Customer:B");
customers.CreateIndex();
customersA.CreateIndex();
customersB.CreateIndex();
customersA.Insert(new Customer { FirstName = "Customer", LastName = "A" });
customersB.Insert(new Customer { FirstName = "Customer", LastName = "B" });

Console.WriteLine("Customers in customers collection");
foreach (var customer in customers)
{
    Console.WriteLine($"Customer: {customer.Id} FirstName: {customer.FirstName}, LastName: {customer.LastName}");
}

Console.WriteLine("Customers in customersA collection");

foreach (var customer in customersA)
{
    Console.WriteLine($"Customer: {customer.Id} FirstName: {customer.FirstName}, LastName: {customer.LastName}");
}

Console.WriteLine("Customers in customersB collection");
            
foreach (var customer in customersB)
{
    Console.WriteLine($"Customer: {customer.Id} FirstName: {customer.FirstName}, LastName: {customer.LastName}");
}
```

the customersA collection will generate this index:

```
"FT.CREATE" "customer-Customer:A-idx" "ON" "Json" "PREFIX" "1" "Customer:A:" "SCHEMA" "$.FirstName" "AS" "FirstName" "TAG" "SEPARATOR" "|" "$.LastName" "AS" "LastName" "TAG" "SEPARATOR" "|" "$.Email" "AS" "Email" "TAG" "SEPARATOR" "|" "$.Age" "AS" "Age" "NUMERIC" "SORTABLE" "$.Home" "AS" "Home" "GEO" "SORTABLE"
```

Whereas customersB will create:

```
"FT.CREATE" "customer-Customer:B-idx" "ON" "Json" "PREFIX" "1" "Customer:B:" "SCHEMA" "$.FirstName" "AS" "FirstName" "TAG" "SEPARATOR" "|" "$.LastName" "AS" "LastName" "TAG" "SEPARATOR" "|" "$.Email" "AS" "Email" "TAG" "SEPARATOR" "|" "$.Age" "AS" "Age" "NUMERIC" "SORTABLE" "$.Home" "AS" "Home" "GEO" "SORTABLE"
```

Then as you insert objects into redis via the prefixed collections, it will produces keys like:

```
"JSON.SET" "Customer:A:01GGCKG1RF0KC6KPT4RGQKQ08M" "." "{\"Id\":\"01GGCKG1RF0KC6KPT4RGQKQ08M\",\"FirstName\":\"Customer\",\"LastName\":\"A\",\"Age\":0,\"Home\":\"0,0\"}"
```
and
```
"JSON.SET" "Customer:B:01GGCKG1RYBSXV89RH82P5RG2F" "." "{\"Id\":\"01GGCKG1RYBSXV89RH82P5RG2F\",\"FirstName\":\"Customer\",\"LastName\":\"B\",\"Age\":0,\"Home\":\"0,0\"}"
```

Because the Prefix for the class `Customer` is `Customer`, the admin collection (just the redis-collection for Customer) will be able to see everything, where as the customersA and customersB collection will only be able to see what they've prefixed to.

As for ACLs, you can of course limit which index is visible to which user, you can also limit keys via prefix I think? (about 85% sure of that lol) so you can manage this by prefix. But of course as long as you are using the right collection for the right customer, the ACLs are really just a backup.

mijail-repower:
Hi everyone, do you know when this new feature will be released? I'm trying to migrate from using `NRediSearch` and `NReJson` to `RedisOM` but this dynamic prefix feature is a blocker.

sidgluckman:
Thank you, this is a great improvement.  Will there be a release of this feature soon?

Issue #249, opened 2022-11-01
Add IRedisConnectionProvider interface


jcreus1:
Hi @slorello89. I'd like to extract an interface for the RedisConnectionProvider class, to allow mocking/substitution for unit testing.

My use case is that I have an ASP.NET Core Web API project injecting a RedisConnectionProvider object into my caching implementation class:

_Program.cs_
![image](https://user-images.githubusercontent.com/30388047/199312048-f4cf1955-7b6b-4d22-97a1-82badc52d7f8.png)

_Caching Implementation:_
![image](https://user-images.githubusercontent.com/30388047/199312325-63bb288a-29ed-4846-b826-cacb5496181d.png)

I am currently unable to unit test this class, as our chosen mocking tool (NSubstitute) requires either an interface or a concrete implementation containing virtual methods it can override. I would therefore like to create an IRedisConnectionProvider interface and to update each reference to RedisConnectionProvider to use this interface. 

Will raise a PR with a simple implementation.

PR #250, opened 2022-11-01
IRedisConnectionProvider interface

jcreus1:
Updates For #249


slorello89 (approved):
This really should have been in from the beginning - thanks for this contribution @jcreus1 - really appreciate it :)

PR #263, opened 2022-11-24
fix:backward-contains-list-type-queries

### LINKED ISSUE #203: Contains numeric not working
I use the test model (Person) and search ages in my list but I get `Could not retrieve value from Value, most likely, it is not properly decorated in the model defining the index.` exception.

The first problem is that, why use nullable int [here](https://github.com/redis/redis-om-dotnet/blob/3eeb031484d5a345ab647926d94dcae888a4a1d0/src/Redis.OM/Common/ExpressionParserUtilities.cs#L304)? 
```csharp
if (resolved is IEnumerable<int> ints)
{
    var sb = new StringBuilder();
    sb.Append('|');
    foreach (var i in ints)
    {
        sb.Append($"[{i} {i}]|");
    }

    sb.Remove(sb.Length - 1, 1);
    return sb.ToString();
}
```

Second problem is my property is nullable however my list is not. So I need to use .Value inside Contains.

And I think last problem in numeric may be query not like that `FT.SEARCH person-idx "@Age:[23 23] | @Age:[25 25]"`

image

```csharp
var people = provider.RedisCollection<Person>();
connection.CreateIndex(typeof(Person));

people.Insert(new Person {
    Name = "Beaker",
    Age = 23,
    Sales = 500000,
    SalesAdjustment = .6,
    Height = 15,
    DepartmentNumber = 3
});

var ages = new List<int> { 23, 25 };

var result = people.Where(w => ages.Contains(w.Age.Value)).ToList();

[Document(StorageType = StorageType.Json, IndexName = "person-idx")]
public class Person {
    [RedisIdField]
    [Indexed]
    [Searchable]

    [Indexed(Sortable = true)]
    public int? Age { get; set; }

    [Indexed(Sortable = true)]
    public double? Height { get; set; }

    [Indexed(Sortable = true)]
    public int? DepartmentNumber { get; set; }

    [Indexed(Sortable = true)]
    public double? Sales { get; set; }

    [Indexed(Sortable = true)]
    public double? SalesAdjustment { get; set; }
}
```

Jeevananthan-23:
Fixes #203, issue with `List` type queries the expression object not parsing inverted contains.

### DIFF
--- a/examples/Redis.OM.FullTextSearch/Program.cs
+++ b/examples/Redis.OM.FullTextSearch/Program.cs
@@ -19,7 +19,7 @@ connection.CreateIndex(typeof(Movie));
 connection.CreateIndex(typeof(Awards));
 
 var redisHelper = new RedisHelper(provider);
-redisHelper.InitializeCustomers();
+redisHelper.InitializeMovies();
 
 //BasicFirstOrDefaultQuery
 // get first or default null value from movies redis limit set to 0 to 1
--- a/examples/Redis.OM.FullTextSearch/RedisHelper/RedisHelper.cs
+++ b/examples/Redis.OM.FullTextSearch/RedisHelper/RedisHelper.cs
@@ -13,7 +13,7 @@ namespace Redis.OM.FullTextSearch.RedisHelper
             movieCollection = provider.RedisCollection<Movie>();
         }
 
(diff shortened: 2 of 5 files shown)


slorello89 (approved):
LGTM 👍

PR #273, opened 2022-12-13
forcing reload on scripting error, using correct sha in some tests

### LINKED ISSUE #272: v0.40 getting NOSCRIPT No matching script. Please use EVAL.
Updated to version 0.4 but getting a lot of NOSCRIPT No matching script. Please use EVAL.... 

`Failed on EVALSHA blah 1 A:B:C:ce32d1e8-9745-ed11-811e-005056a6c176 SET $.IsOnline true SET $.FristAtUTC 1670907631285 SET $.LastAtUTC 1670907334327`

`Failed on EVALSHA 662066e89db469f8f9a8a7409b603b218bc4eac8 1 A:B:C:D@google.com:f6e52183-930c-ed11-8106-005056a63161 {"Id":"D@google.com:f6e52183-930c-ed11-8106-005056a63161","MemberId":"f6e52183-930c-ed11-8106-005056a63161","Recipient":"D@google.com","GroupId":"f5e52183-930c-ed11-8106-005056a63161","Type":123,"Count":1}`

slorello89:
Fixes #272 - issue is when redis is rolled over, the client's not properly handling the reload and trying to send `EVALSHA`'s when it really needs to either run an `EVAL` or script reload. Because At this point in the error flow we are already one or two round trips to redis in, we'll use an `EVAL` for the initial condition, and empty out our collection of SHA's so that we force the client to reload all scripts it was previously running

### DIFF
--- a/src/Redis.OM/Redis.OM.csproj
+++ b/src/Redis.OM/Redis.OM.csproj
@@ -6,9 +6,9 @@
     <RootNamespace>Redis.OM</RootNamespace>
     <Nullable>enable</Nullable>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <PackageVersion>0.4.0</PackageVersion>
-    <Version>0.4.0</Version>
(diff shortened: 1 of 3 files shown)


shacharPash (approved):
LGTM :+1:

PR #277, opened 2022-12-16
issue with closures messing up contains

### LINKED ISSUE #276: Issue with Contains in the latest version 0.4.0
Hello!

After last update I faced with issue in Contains().

Something like that works fine:
`var cacheDtos = collection.Where(x => x.Tags.Contains("test")).ToList();`

but when I use variable it throw exception "Called contains for a non-indexed property"
```
var tag = "test";
var cacheDtos = collection.Where(x => x.Tags.Contains(tag)).ToList();
```

Property Tags has necessary attribute 
    
```
    [Indexed]
    public List<string> Tags { get; set; }
```
I've tried to use another attribute Searchable, but it doesn't work too.

Thank you!

slorello89:
Fixes #276


slorello89:
@shacharPash - can one of you have a look when you have a chance?

shacharPash (approved):
LGTM :+1:

PR #278, opened 2022-12-16
feature: Bulk Operations 

### LINKED ISSUE #261: Bulk Operations
Hello, Adding the ability to do bulk operations like InsertMany and deleteMany, and updateMany.

### LINKED ISSUE #274: Bulk Operations batch
hi @slorello89, I would like to work on this enhancement. As I already added a batch by #246 
Few items for clarification:
- **InsertMany** which is already added but anyway this function is `Async` so I should refactor and add a `Sync` function for the bulk insert operation -  please add your points.
-  **` WhenKey`** - should be part of the bulk operation?
- **UpdateMany** - as you mentioned above this function should check for the snapshot to get diff and update the record accordingly.
- **DelMany** - uses `Unlink` to remove the enumerated item.(TC: O(N)) iterator into each key and take a round trip also expensive --> adding your suggestion.
- **All the bulk operations should have `Sync` & `Async` methods?**

_Originally posted by @Jeevananthan-23 in https://github.com/redis/redis-om-dotnet/issues/261#issuecomment-1350962472_

Jeevananthan-23:
**Adding feature to Redis-om-dotnet where developers can perform Bulk Operations like Insert, Delete, and Update.**
- [x] Bulk Insert - Insertion of records supports Hash and JSON docs(performance improvement `1000` records in an avg of `200ms`) Closes https://github.com/Jeevananthan-23/redis-om-dotnet/issues/1#issue-1496163279

- [x] Bulk Delete - Unlinks the keys from the Redis.  Closes https://github.com/Jeevananthan-23/redis-om-dotnet/issues/3#issue-1496176859

- [x]  Bulk Update - Update the records.  Closes https://github.com/Jeevananthan-23/redis-om-dotnet/issues/2#issue-1496170149

- [ ] Closes #274, Closes #261, Closes https://github.com/Jeevananthan-23/redis-om-dotnet/issues/4#issue-1496189258 🎉

### DIFF
--- a/src/Redis.OM/Searching/IRedisCollection.cs
+++ b/src/Redis.OM/Searching/IRedisCollection.cs
@@ -164,11 +164,24 @@ namespace Redis.OM.Searching
         void Delete(T item);
 
         /// <summary>
+        /// Deletes the List of items from Redis.
+        /// </summary>
+        /// <param name="items">The items to be deleted.</param>
+        void Delete(IEnumerable<T> items);
+
+        /// <summary>
         /// Deletes the item from Redis.
         /// </summary>
         /// <param name="item">The item to be deleted.</param>
         /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
         Task DeleteAsync(T item);
+
+        /// <summary>
+        /// Deletes the List of items from Redis.
+        /// </summary>
+        /// <param name="items">The items to be deleted.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        Task DeleteAsync(IEnumerable<T> items);
 
         /// <summary>
         /// Async method for enumerating the collection to a list.
--- a/src/Redis.OM/Searching/IRedisCollection.cs
+++ b/src/Redis.OM/Searching/IRedisCollection.cs
@@ -151,11 +151,24 @@ namespace Redis.OM.Searching
         void Update(T item);
 
         /// <summary>
+        /// Updates the provided items in Redis. Document must have a property marked with the <see cref="RedisIdFieldAttribute"/>.
+        /// </summary>
+        /// <param name="items">The items to update.</param>
+        void Update(IEnumerable<T> items);
+
+        /// <summary>
         /// Updates the provided item in Redis. Document must have a property marked with the <see cref="RedisIdFieldAttribute"/>.
         /// </summary>
         /// <param name="item">The item to update.</param>
         /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
         Task UpdateAsync(T item);
+
+        /// <summary>
+        /// Updates the provided items in Redis. Document must have a property marked with the <see cref="RedisIdFieldAttribute"/>.
+        /// </summary>
+        /// <param name="items">The items to update.</param>
+        /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+        ValueTask UpdateAsync(IEnumerable<T> items);
 
         /// <summary>
         /// Deletes the item from Redis.
--- a/src/Redis.OM/Searching/IRedisCollection.cs
+++ b/src/Redis.OM/Searching/IRedisCollection.cs
@@ -92,22 +92,7 @@ namespace Redis.OM.Searching
         /// </summary>
         /// <param name="items">The items to insert.</param>
         /// <returns>The list of Keys.</returns>
-        List<string> Insert(IEnumerable<T> items);
-
-        /// <summary>
-        /// Inserts list of items into redis.
-        /// </summary>
-        /// <param name="items">The items to insert.</param>
-        /// <param name="timeSpan">The expiration time of the document (TTL).</param>
-        /// <returns>The list of Keys.</returns>
-        List<string> Insert(IEnumerable<T> items, TimeSpan timeSpan);
-
-        /// <summary>
-        /// Inserts list of items into redis.
-        /// </summary>
-        /// <param name="items">The items to insert.</param>
-        /// <returns>The list of Keys.</returns>
-        Task<List<string>> InsertAsync(IEnumerable<T> items);
+        Task<List<string>> Insert(IEnumerable<T> items);
 
         /// <summary>
         /// Inserts list of items into redis.
@@ -115,7 +100,7 @@ namespace Redis.OM.Searching
         /// <param name="items">The items to insert.</param>
         /// <param name="timeSpan">The timespan of the document's (TTL).</param>
         /// /// <returns>The list of Keys.</returns>
-        Task<List<string>> InsertAsync(IEnumerable<T> items, TimeSpan timeSpan);
+        Task<List<string>> Insert(IEnumerable<T> items, TimeSpan timeSpan);
 
         /// <summary>
         /// finds an item by it's ID or keyname.
@@ -149,12 +134,6 @@ namespace Redis.OM.Searching
         /// </summary>
         /// <param name="item">The item to update.</param>
         void Update(T item);
-
-        /// <summary>
-        /// Updates the provided items in Redis. Document must have a property marked with the <see cref="RedisIdFieldAttribute"/>.
-        /// </summary>
-        /// <param name="items">The items to update.</param>
-        void Update(IEnumerable<T> items);
 
         /// <summary>
         /// Updates the provided item in Redis. Document must have a property marked with the <see cref="RedisIdFieldAttribute"/>.
--- a/src/Redis.OM/Searching/IRedisCollection.cs
+++ b/src/Redis.OM/Searching/IRedisCollection.cs
@@ -88,6 +88,36 @@ namespace Redis.OM.Searching
         string? Insert(T item, WhenKey when, TimeSpan? timeSpan = null);
 
         /// <summary>
+        /// Inserts list of items into redis.
+        /// </summary>
+        /// <param name="items">The items to insert.</param>
+        /// <returns>The list of Keys.</returns>
+        List<string> Insert(IEnumerable<T> items);
+
+        /// <summary>
+        /// Inserts list of items into redis.
+        /// </summary>
+        /// <param name="items">The items to insert.</param>
+        /// <param name="timeSpan">The expiration time of the document (TTL).</param>
+        /// <returns>The list of Keys.</returns>
+        List<string> Insert(IEnumerable<T> items, TimeSpan timeSpan);
+
+        /// <summary>
+        /// Inserts list of items into redis.
+        /// </summary>
+        /// <param name="items">The items to insert.</param>
+        /// <returns>The list of Keys.</returns>
+        Task<List<string>> InsertAsync(IEnumerable<T> items);
+
+        /// <summary>
+        /// Inserts list of items into redis.
+        /// </summary>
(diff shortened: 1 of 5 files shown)


VagyokC4:
> * [x]  Bulk Insert - Insertion of records supports Hash and JSON docs(performance improvement `1000` records in an avg of `200ms`)

Question: Are these records added to the index, i.e. is this a real bulk import, or will these bulk records get indexed no problem?

slorello89:
@VagyokC4 - If the index is already established, indexing them is part of the insertion, it happens synchronously and automatically, if the index has yet to be created, when the index IS created everything is indexed in the background.

VagyokC4:
> @VagyokC4 - If the index is already established, indexing them is part of the insertion, it happens synchronously and automatically, if the index has yet to be created, when the index IS created everything is indexed in the background.

I've seen instances where I've had the index established, but I do a bulk insert, and those records are not in the index or searchable. I am forced to re-index the collection.

Jeevananthan-23:
> > @VagyokC4 - If the index is already established, indexing them is part of the insertion, it happens synchronously and automatically, if the index has yet to be created when the index IS created everything is indexed in the background.
> 
> I've seen instances where I've had the index established, but I do a bulk insert, and those records are not in the index or searchable. I am forced to re-index the collection.

I think your uses case could be as follows : 
```cs
        [Fact]
        public async Task TestBulkInsertAsync()
        {
            var collection = new RedisCollection<Person>(_connection);
            var persons = new List<Person>() {
                new Person() { Name = "Alice", Age = 51, NickNames = new[] { "Ally", "Alie", "Al" } },
                new Person() { Name = "Robert", Age = 37, NickNames = new[] { "Bobby", "Rob", "Bob" } },
                new Person() { Name = "Jeeva", Age = 22, NickNames = new[] { "Jee", "Jeev", "J" } },
                new Person() { Name = "Martin", Age = 60, NickNames = new[] { "Mart", "Mat", "tin" } }
                };
            var keys = await collection.InsertAsync(persons); // performs JSON.SET create keys and emit the list of keys 

            var persons2 = new List<Person>() {
                new Person() { Name = "Alice", Age = 51, NickNames = new[] { "Ally", "Alie", "Al" }, IsEngineer = true },
                new Person() { Name = "Robert", Age = 37, NickNames = new[] { "Bobby", "Rob", "Bob" }, IsEngineer = false },
                new Person() { Name = "Jeeva", Age = 22, NickNames = new[] { "Jee", "Jeev", "J" }, DepartmentNumber = 201 },
                new Person() { Name = "Martin", Age = 60, NickNames = new[] { "Mart", "Mat", "tin" }, TagField = "Martin" }
                };

            var keys2 = await collection.InsertAsync(persons2); // create keys and emit the list of keys

            var people = collection.Where(x => x.NickNames.Contains("Bob") || x.NickNames.Contains("Alie")).ToList();
            Assert.Contains(people, x => x.Name == persons.First().Name);
            Assert.NotEqual(keys, keys2); // not performs any re-indexing because the keys are not the same.
        }
```

**The above test will answer your question and your intention here is to  perform the update the records so use UpdateMany in this case, Thanks!**

slorello89:
> > @VagyokC4 - If the index is already established, indexing them is part of the insertion, it happens synchronously and automatically, if the index has yet to be created, when the index IS created everything is indexed in the background.
> 
> I've seen instances where I've had the index established, but I do a bulk insert, and those records are not in the index or searchable. I am forced to re-index the collection.

This sounds like it could be an indexing issue (ie your old index has something incompatible with the data in your index) - you can check to see if there were any indexing failures with FT.INFO

Jeevananthan-23:
Hi @slorello89, looking forward to your feedback on this PR and will make changes accordingly. Thanks!

src/Redis.OM/Searching/IRedisCollection.cs
```diff
@@ -120,26 +150,52 @@ public interface IRedisCollection<T> : IOrderedQueryable<T>, IAsyncEnumerable<T>
         /// <param name="item">The item to update.</param>
         void Update(T item);
 
+        /// <summary>
+        /// Updates the provided items in Redis. Document must have a property marked with the <see cref="RedisIdFieldAttribute"/>.
+        /// </summary>
+        /// <param name="items">The items to update.</param>
+        void Update(IEnumerable<T> items);
```
slorello89, line 157: IMO Sync Update is a bad idea, I wrote `Save` early on in Redis OM just in case someone needed something like this, but I don't think this should be propagated, as this is going to incur many round trips to Redis, I would just axe this and keep the Async (named `UpdateAsync`)

src/Redis.OM/Searching/IRedisCollection.cs
```diff
@@ -87,6 +87,36 @@ public interface IRedisCollection<T> : IOrderedQueryable<T>, IAsyncEnumerable<T>
         /// <returns>the Id of the newly inserted item, or null if not inserted.</returns>
         string? Insert(T item, WhenKey when, TimeSpan? timeSpan = null);
 
+        /// <summary>
+        /// Inserts list of items into redis.
+        /// </summary>
+        /// <param name="items">The items to insert.</param>
+        /// <returns>The list of Keys.</returns>
+        List<string> Insert(IEnumerable<T> items);
```
slorello89, line 95: Ditto for these inserts, it's going to incur a ton of round trips to Redis so it's not really appropriate to have here. Also since you are changing the signature of `Insert` this would be a breaking change. Revert all the stuff for the bulk inserts, we'll have to live with the weird `Insert` vs `InsertAsync` name. Given that there's only one multi-variat insert I think that's fine.

slorello89 (changes requested):

src/Redis.OM/Searching/IRedisCollection.cs
Jeevananthan-23: JSYK I'm using the [pipeline concept](https://stackexchange.github.io/StackExchange.Redis/PipelinesMultiplexers#pipelining) to get all requests onto the network immediately, eliminating most of the latency and reducing packets send to Redis. (Test is done for Sync operation and improved performance)

BTW adding some points for some improvements in the code 
* Adding `ArgumentNullException` for all the methods 
```csharp
  public void Update(IEnumerable<T> items)
        {
            if (items == null)
            {
                throw new ArgumentNullException("items cannot be null. Please provide a non-null list of items.");
            }
}
```
* Naming conversions for Bulk operation methods
```
BulkInsert()
BulkUpdate() or UpdateMany()
BulkDelete() or DeleteMany()
```
*Later I'll work on adding doc for all the bulk operations*

Thanks!

slorello89: FYI @Jeevananthan-23 - the way you are trying to do this will cause deadlocks - so not a great idea - `Task.Wait` is fine if you have a very defined set of threads you are working with and you KNOW that your waits aren't going to seize all the threads that you need to run the tasks, but if you try to do this in a thread constrained environment (e.g. imagine your application only had 1 thread), the `Task.Wait`'s you'll accumulate can stop all the threads you need to run your tasks and deadlock your application. You might be interested to see how it's referred to in [StackExchange.Redis](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/PhysicalConnection.cs#LL971C17-L971C17) - but they take great pains to make sure their tasks in a detached context that will not be messed with, we make no such promises.

Jeevananthan-23: @slorello89 I'll dive deep into it and make some test analysis through it. If possible guide me in this implementation to come up with a better solution.

slorello89: As I mentioned - there isn't a good way to do this synchronously, so I'd just as soon remove the sync update/insert, can keep the sync delete because that's purely variadic.

Jeevananthan-23: > BTW adding some points for some improvements in the code
> 
> * Adding `ArgumentNullException` for all the methods
> 
> ```cs
>   public void Update(IEnumerable<T> items)
>         {
>             if (items == null)
>             {
>                 throw new ArgumentNullException("items cannot be null. Please provide a non-null list of items.");
>             }
> }
> ```
> 
> * Naming conversions for Bulk operation methods
> 
> ```
> BulkInsert()
> BulkUpdate() or UpdateMany()
> BulkDelete() or DeleteMany()
> ```
> 
> _Later I'll work on adding doc for all the bulk operations_
> 
> Thanks!
@slorello89 if there is no way of doing it then fine. Any commands on those above points ?

Jeevananthan-23: > FYI @Jeevananthan-23 - the way you are trying to do this will cause deadlocks - so not a great idea - `Task.Wait` is fine if you have a very defined set of threads you are working with and you KNOW that your waits aren't going to seize all the threads that you need to run the tasks, but if you try to do this in a thread constrained environment (e.g. imagine your application only had 1 thread), the `Task.Wait`'s you'll accumulate can stop all the threads you need to run your tasks and deadlock your application. You might be interested to see how it's referred to in [StackExchange.Redis](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/PhysicalConnection.cs#LL971C17-L971C17) - but they take great pains to make sure their tasks are in a detached context that will not be messed with, we make no such promises.

 `sync` makes the thread wait to complete, if it `multithreaded` application this is blocking call and user should aware of this process and go with `async`. As you mentioned before the `Save` is not used by over `SaveAsync` yes it's true but [ef core implementation also similar](https://github.com/dotnet/EntityFramework.Docs/blob/main/entity-framework/core/providers/cosmos/limitations.md#azure-cosmos-db-sdk-limitations) so it's upto the user should understand the limitations to prevent deadlocks.

Jeevananthan-23: @slorello89  adding one more solution to this issue using `GetAwaiter` instant of `Wait` may still cause deadlocks if it is used improperly. 

```csharp

public List<string> Insert(IEnumerable<T> items)
        {
            var tasks = new List<Task<string>>();
            foreach (var item in items.Distinct())
            {
                tasks.Add(((RedisQueryProvider)Provider).Connection.SetAsync(item));
            }

            Task.WhenAll(tasks).GetAwaiter();
            var result = tasks.Select(x => x.Result).ToList();
            return result;
        }

// calling Async but GetResult() is block the current thread
 public List<string> Insert(IEnumerable<T> items)
        {
            return InsertAsync(items).GetAwaiter().GetResult();
        }
```

slorello89: @Jeevananthan-23 - fundamentally the issue is that running Async code from Sync code is an anti-pattern, and any properly sync patterns for this would be horrible on performance. This is why my conclusion has just been to remove the sync methods.

`InsertAsync`
`UpdateAsync`
`Delete`
`DeleteAsync`

Would be the names I would prefer, but since `Insert` is already the name, we ought to keep that (otherwise it will be a breaking change for people already using this)

Jeevananthan-23: @slorello89 - Thanks for bearing with me a long conversation and pushing the feedback changes.

slorello89 (approved):
LGTM 👍

Issue #275, opened 2022-12-15
When using a large Or operator, Syntax error  is thrown...


zulander1:
I have an issue when query with a large Or (like 100 or conditions) operator I get a syntax error, I am not sure if this RedisSearch or Redis-OM, i've changed the limit to 200 I am using ExpressionVisitor to convert "Contain":

```
var provider = new RedisConnectionProvider("redis://localhost:6379");
var connection = provider.Connection;
var customers = provider.RedisCollection<Customer>();
var customerAggregations = provider.AggregationSet<Customer>();

// Create index
connection.CreateIndex(typeof(Customer));

// Get Index info
var indexinfo = connection.GetIndexInfo(typeof(Customer));

// Insert Object
var listOfCustomers = Enumerable.Range(1, 100).Select(x => new Customer() { Id = Guid.NewGuid(), LastName = x.ToString() });
foreach (var item in listOfCustomers)
{
    await customers.InsertAsync(item);
}

var ids = listOfCustomers.Select(x => x.Id).Take(100).ToArray();

Expression<Func<Customer, bool>> expression = x => ids.Contains(x.Id);
RedisQueryVisitor redisQueryVisitor = new();
var visited = (Expression<Func<Customer, bool>>)redisQueryVisitor.Visit(expression);

var list = customers.Where(visited).Take(200).ToList();

Console.ReadLine();

[Document(StorageType = StorageType.Json)]
public class Customer
{
    [Indexed][RedisIdField] public Guid Id { get; set; }
    [Indexed] public string LastName { get; set; }
}

public class RedisQueryVisitor : ExpressionVisitor
{
    protected override Expression VisitMethodCall(MethodCallExpression node)
    {
        if (node.Method.Name == "Contains" && node.Arguments.Count == 2)
        {
            var right = (MemberExpression)node.Arguments[1];
            var left = (MemberExpression)node.Arguments[0];

            if (left.Member.MemberType == MemberTypes.Property)
            {
                PropertyInfo leftPI = (PropertyInfo)left.Member;

                //Check if the secound parameter is collection
                if (leftPI.PropertyType.IsArray)
                    return base.VisitMethodCall(node);
            }

            #region Get List

            var list = (IEnumerable)Expression.Lambda<Func<object>>(node.Arguments[0]).Compile()();

            var castedOBject = list.OfType<object>().ToList();

            #endregion Get List

            if (castedOBject.Any())
[...]

PR #279, opened 2022-12-17
Fixed contains for guid/ulid/enum

zulander1:
this is to fix issue https://github.com/redis/redis-om-dotnet/issues/275.


slorello89 (approved):
Good catch @zulander1 - I expanded the scope a bit to encompass ULIDs and enums, I think this is good to go - thanks for submitting.

PR #285, opened 2023-01-07
support for JSON.ArrIndex with none scalar value

### LINKED ISSUE #171: Issue with Updating collections with new objects
There seems to be a weird issue with how diffs are being built when new objects are being added to arrays withing JSON objects, it seems like the nodes are not being created correctly.

### LINKED ISSUE #269: When saving a nested item using SaveAsync (with state tracking error) getting an error
We are using Redis JSON (moved from Hash to JSON model recently) and need to update a child record field. However, when the child record is modified and saved, there is an error during SaveAsync.

StackExchange.Redis.RedisServerException: 'ERR Error running script (call to f_cbbf1c4fab5064f419e469cc51c563f8bf51e6fb): @user_script:5: WRONGTYPE wrong type of path value - expected scalar but found {    "Id": "item1",    "Name": "item1"  }'

Below is a repro code using top-level statements. Error is thrown on Save/SaveAsync call. Only a single property has to be updated in the child record, is there any other way to do this? Direct JSONSet might help too, but wanted to avoid that when possible.

`
using Redis.OM;
using Redis.OM.Modeling;
using StackExchange.Redis;

var cm = ConnectionMultiplexer.Connect("host.docker.internal:6379");
var provider = new RedisConnectionProvider(cm);

var w1 = new Item { Id = "item1", Name = "item1", };
var w2 = new Item { Id = "item2", Name = "item2", };

var o1 = new OrderTest
{
	Id = "111",
	Items = new List<Item> { w1, w2 },
};

provider.RedisCollection<OrderTest>().Insert(o1);

var orders = provider.RedisCollection<OrderTest>(true);
var o2 = orders.FindById(o1.Id);
o2.Items[0].Name = o2.Items[0].Name + ".1";

orders.Save(); // throws error

//-------------- POCO
[Document(StorageType = StorageType.Json, Prefixes = new[] { "OrderTest" })]
public sealed partial class OrderTest
{
	[RedisIdField]

	[RedisField]
	[Indexed(CascadeDepth = 2)]
	public List<Item> Items { get; set; } = new();
}

public sealed partial class Item
{
	[RedisField]

	[RedisField]
	public string? Name { get; set; }
}

`

Jeevananthan-23:
resolved #269
resolved #171

### DIFF
--- a/test/Redis.OM.Unit.Tests/RediSearchTests/ObjectWithEmbeddedArrayOfObjects.cs
+++ b/test/Redis.OM.Unit.Tests/RediSearchTests/ObjectWithEmbeddedArrayOfObjects.cs
@@ -6,6 +6,9 @@ namespace Redis.OM.Unit.Tests.RediSearchTests
     [Document(StorageType = StorageType.Json)]
     public class ObjectWithEmbeddedArrayOfObjects
     {
+        [RedisIdField]
+        public string Id { get; set; }
+
         [Indexed(JsonPath = "$.City")]
         [Indexed(JsonPath = "$.State")]
         [Indexed(JsonPath = "$.AddressType")]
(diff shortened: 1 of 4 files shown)


Jeevananthan-23:
Hello @slorello89, still the `EVALSHA` #273 fail to run for some test for more details please check the build.

Thanks!

slorello89 (approved):
LGTM 👍 thanks @Jeevananthan-23

PR #295, opened 2023-02-02
Fixing races

slorello89:
There's some lurking race's going on within Redis OM, particularly when it comes to highly parallelized async operations (e.g. bulk update / fetch).

Two primary culprits:

1. The dictionary we're using to hold the scripts sha's are not thread-safe, so if the sha collection was updated while also being read, the dictionary was throwing an invalid operation exception.
2. The Bulk Update/FindById async methods both leverage their respective non-bulk versions. Which would be fine if not for the fact that they both update the State Manager, since the state manager is not meant to be thread-safe, when multiple async tasks hit it simultaneously, all sorts of unpredictable things can happen. 

This PR addresses both of these issue.


bsbodden (approved):
LGTM!

Issue #286, opened 2023-01-09
[RedisIdField] issue when using type different then string, GUID or Ulid 


Zlejsi:
When using different type then string, GUID or Ulid at [RedisIdField] attribute, program throw an exception.
```C# System.ArgumentException: 'Object of type 'System.String' cannot be converted to type 'System.Int32'.'```
As example i used System.Int32 as type.

Problem is at converting Id generated by GenerateId() method.

```class RedisObjectHandler.cs```
```C#
internal static string SetId(this object obj)
    {
      Type type = obj.GetType();
      DocumentAttribute customAttribute = Attribute.GetCustomAttribute((MemberInfo) type, typeof (DocumentAttribute)) as DocumentAttribute;
      PropertyInfo propertyInfo = ((IEnumerable<PropertyInfo>) type.GetProperties()).FirstOrDefault<PropertyInfo>((Func<PropertyInfo, bool>) (x => Attribute.GetCustomAttribute((MemberInfo) x, typeof (RedisIdFieldAttribute)) != null));
      string str = customAttribute != null ? customAttribute.IdGenerationStrategy.GenerateId() : throw new MissingMemberException("Missing Document Attribute decoration");
      if (propertyInfo != (PropertyInfo) null)
      {
        Type propertyType = propertyInfo.PropertyType;
        if (!((IEnumerable<Type>) new Type[3]
        {
          typeof (string),
          typeof (Guid),
          typeof (Ulid)
        }).Contains<Type>(propertyType) && !propertyType.IsValueType)
          throw new InvalidOperationException("Software Defined Ids on objects must either be a string, ULID, Guid, or some other value type.");
        object obj1 = propertyInfo.GetValue(obj);
        if (!RedisObjectHandler.TypeDefaultCache.ContainsKey(propertyType))
          RedisObjectHandler.TypeDefaultCache.Add(propertyType, Activator.CreateInstance(propertyType));
        if (obj1?.ToString() != RedisObjectHandler.TypeDefaultCache[propertyType]?.ToString())
          str = propertyInfo.GetValue(obj).ToString();
        else if (propertyType == typeof (Guid))
          propertyInfo.SetValue(obj, (object) Guid.Parse(str));
        else if (propertyType == typeof (Ulid))
          propertyInfo.SetValue(obj, (object) Ulid.Parse(str));
        else
         propertyInfo.SetValue(obj, (object) str);
      }
[...]

PR #301, opened 2023-02-05
Fixed cast to different type

zulander1:
Fixed https://github.com/redis/redis-om-dotnet/issues/286

### DIFF
--- a/src/Redis.OM/Common/ExpressionTranslator.cs
+++ b/src/Redis.OM/Common/ExpressionTranslator.cs
@@ -375,7 +375,7 @@ namespace Redis.OM.Common
 
             if (exp is MemberExpression member)
             {
-                return $"{member.Member.Name}";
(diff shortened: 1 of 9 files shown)


zulander1:
@slorello89 not sure why failing.. test are passing

JuliusMikkela:
I've noticed that the tests on main are a bit unstable, both locally and in the ci, with varying tests of the Search and SearchFunctional tests failing because the StateManager gets a race condition or certain searches are apparently never executed. This only seems to happen when running the entire collections or at least multiple tests, as if you re-run the tests individually they always succeed. I haven't had time to figure out exactly why this happens.

slorello89:
The races should be mostly fixed now see #295
I think this is a timing issue in the tests that were introduced with the `When` functionality (definitely not related to this PR)

Issue #213, opened 2022-09-25
Updating a collection that has been initialized with null is not working 


zulander1:
Spend a lot of time figuring this one out. This code does not update the collection if the property is null at insert,  if the property is empty (i.e.: Array.Empty<string>()) it will work.

```

var bob = new Customer
{
    Id = Guid.Parse("9f286de2-10ae-4366-aa35-d582b64f6f4d"),
    Age = 35,
    FirstName = "Bob",
    LastName = "Smith",
};
var customerOne = provider.RedisCollection<Customer>();
 
await customerOne.InsertAsync(bob);
var test1 = await customerOne.Where(x => x.Age == 35).ToListAsync(); // this is to have it in the snapshot!

bob.Tags = new[] { "some tag" };
await customerOne.UpdateAsync(bob);


[Document(StorageType = StorageType.Json)]
public class Customer
{
    [RedisIdField] public Guid Id { get; set; }
    [Indexed] public string FirstName { get; set; }
    [Indexed] public string[] Tags { get; set; }
    [Indexed] public string LastName { get; set; }
    [Indexed] public string Email { get; set; }
    [Indexed] public int Age { get; set; }
}
```

zulander1:
@slorello89 hey do you think we fix this for the next release? I looked at it and I think we need to include JSON.ARRINSERT in the Lua script

PR #302, opened 2023-02-05
Update on null collection

zulander1:
This should fix issue https://github.com/redis/redis-om-dotnet/issues/213


slorello89 (approved):
LGTM 👍 thanks for this @zulander1!

PR #306, opened 2023-02-07
integration test timing issues

slorello89:


slorello89:
Had to resolve a couple of timing based issues with the integration tests for Redis OM. Two primary problems

1: In a previous PR I fixed an issue where if a script was reloaded properly, the mock sha that we use for our unit tests was wrong, it appears that I missed a couple of them
2: The timing from #206 was slightly off - yes the "we started out with 5000 ms timeout, delayed a second so we should have less than a 4000 ms timeout" logic seems sound, but `Task.Delay` only guarantees accuracy to the resolution of the system clock, so it was a few millisecond off, I added some extra buffer so hopefully that shouldn't be an issue anymore

I've run the tests 10 times without a failure (was failing at least 1 / 3 times before) so hopefully this fixes it 🤞

bsbodden (approved):
LGTM!

Issue #329, opened 2023-03-08
Create index with custom Prefixes missing ":"


29654761:
Create index with custom Prefixes missing ":" 
It's lead to search out rows from other tables witch names same Prefixes key.
If add ":" in DocumentAttribute.Prefixes , this will be the key format like key::id (double ":")

slorello89:
Hi @29654761 - that's by design, the typical namespace pattern in Redis is divided by a single `:`, the Prefix is used for two things 

1. Scoping the domain of the index (so any keys that start with the prefix are indexed)
2. When the keynames are generated at insertion by Redis OM, it DOES add an `:` between the prefix and the Id.

Not sure what the general purpose of having a double `:` is? If it's really needed, you can always add it to the prefix.

29654761:
> Hi @29654761 - that's by design, the typical namespace pattern in Redis is divided by a single `:`, the Prefix is used for two things
> 
> 1. Scoping the domain of the index (so any keys that start with the prefix are indexed)
> 2. When the keynames are generated at insertion by Redis OM, it DOES add an `:` between the prefix and the Id.
> 
> Not sure what the general purpose of having a double `:` is? If it's really needed, you can always add it to the prefix.

The problem is:  
add colon to DocumentAttribute.Prefixes, The key will be key:::id.
If don't add colon to DocumentAttribute.Prefixes, The index will no colon, it lead to search out wrong results

slorello89:
@29654761 Can you provide a more concrete example of this? I'm having a hard time groking what you're talking about here or why there would be an issue - your prefix should just be the prefix - it's not a required field because Redis OM will just use whatever the fully-qualified class-name is without a prefix, and if you do add a prefix - then it will just use that, and when it creates the key - prepend a colon to the Id that it uses.

29654761:
> @29654761 Can you provide a more concrete example of this? I'm having a hard time groking what you're talking about here or why there would be an issue - your prefix should just be the prefix - it's not a required field because Redis OM will just use whatever the fully-qualified class-name is without a prefix, and if you do add a prefix - then it will just use that, and when it creates the key - prepend a colon to the Id that it uses.

HI @slorello89 
Sorry for my poor English.
for example:
There are two tables:
```code
[Document(StorageType = StorageType.Json, Prefixes = new[] { "persion" })]
public class Persion
{
        [RedisIdField]
        public Ulid id { get; set; }
        [Indexed]
        public string name { get; set; }
        ...
}

[...]

PR #330, opened 2023-03-08
Fix creating index with custum Prefixes missing colon (#329)

29654761:


slorello89:
This would be a break - and the issue here is mostly cosmetic as far as I can tell, so closing this out - thanks for opening though!

PR #338, opened 2023-03-25
Async for bulk operation

zulander1:
Added Async suffix on async operations

### DIFF
--- a/src/Redis.OM/Searching/RedisCollection.cs
+++ b/src/Redis.OM/Searching/RedisCollection.cs
@@ -695,6 +695,25 @@ namespace Redis.OM.Searching
         }
 
         /// <inheritdoc/>
+        public async Task<List<string?>> InsertAsync(IEnumerable<T> items, WhenKey when, TimeSpan? timeSpan = null)
+        {
+            var distinct = items.Distinct().ToArray();
+            if (!distinct.Any())
+            {
+                return new List<string?>();
+            }
+
+            var tasks = new List<Task<string?>>();
+            foreach (var item in distinct)
+            {
+                tasks.Add(((RedisQueryProvider)Provider).Connection.SetAsync(item, when, timeSpan));
+            }
+
+            var result = await Task.WhenAll(tasks);
+            return result.ToList();
+        }
+
+        /// <inheritdoc/>
         public T? FindById(string id)
         {
             var prefix = typeof(T).GetKeyPrefix();
--- a/src/Redis.OM/Searching/RedisCollection.cs
+++ b/src/Redis.OM/Searching/RedisCollection.cs
@@ -476,7 +476,11 @@ namespace Redis.OM.Searching
             query.Limit = new SearchLimit { Number = 1, Offset = 0 };
             var res = _connection.Search<T>(query);
             var result = res.Documents.FirstOrDefault();
-            SaveToStateManager(result.Key, result.Value);
+            if (result.Key != null)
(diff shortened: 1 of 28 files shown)


src/Redis.OM/Searching/RedisCollection.cs
```diff
@@ -694,6 +694,25 @@ public async Task<List<string>> Insert(IEnumerable<T> items, TimeSpan timeSpan)
             return result.ToList();
         }
 
+        /// <inheritdoc/>
+        public async Task<List<string?>> InsertAsync(IEnumerable<T> items, WhenKey when, TimeSpan? timeSpan = null)
```
slorello89, line 698: @zulander1 - what's the purpose of having the `List<string?>` version? A failure to create the document in Redis?

zulander1: None it was already there didn't want to change it ! I guess we can remove it ?

slorello89: yeah, pull it out I think (I don't think this one was actually here - it could just be the diff is acting weird?)

zulander1: so this https://github.com/redis/redis-om-dotnet/blob/b95ac2244ad5377fdf813564ca2b658e1e6c882b/src/Redis.OM/Searching/RedisCollection.cs#L648 and and https://github.com/redis/redis-om-dotnet/blob/b95ac2244ad5377fdf813564ca2b658e1e6c882b/src/Redis.OM/Searching/RedisCollection.cs#L679 are returning a list, however I am going to remove it ! thanks'

zulander1: Removed

slorello89: err I think you might of pushed to the wrong branch?

zulander1: Yeah you're right! I'll fix it this afternoon

zulander1: fixed !

src/Redis.OM/Searching/RedisCollection.cs
```diff
@@ -690,8 +690,25 @@ public async Task<List<string>> Insert(IEnumerable<T> items, TimeSpan timeSpan)
                 tasks.Add(((RedisQueryProvider)Provider).Connection.SetAsync(item, timeSpan));
             }
 
-            var result = await Task.WhenAll(tasks);
-            return result.ToList();
+            await Task.WhenAll(tasks);
+        }
+
+        /// <inheritdoc/>
+        public async Task InsertAsync(IEnumerable<T> items, WhenKey when, TimeSpan? timeSpan = null)
```
slorello89, line 697: Ok I see whats happening you are adding one with the `WhenKey` condition - I'm sorry I didn't catch that before. You were right! This needs to return a `List<string?>` because yes - there are cases where it would not have created anything - sorry about that - yeah just return the `List<string?>`

slorello89 (changes requested):

src/Redis.OM/Searching/RedisCollection.cs
zulander1: okay I've deleted the last commit, it should be back to the original

slorello89 (approved):
LGTM 👍

PR #344, opened 2023-03-29
Cannot translate complex object filter on the right of where clause in aggregation sets 

imansafari1991:


slorello89:
Not atm @imansafari1991 - sorry I've been fairly busy and haven't had the bandwidth to get to this.

Issue #352, opened 2023-04-14
Can Redis OM support RESTful Webdis interface to avoid persistent connection issues?


VagyokC4:
Hi @slorello89.  Hope all is well.  We're working on a hybrid solution that uses both a cloud Redis instance and an on-prem Redis instance. Due to the persistent connection nature of Redis, if the on-prem instance becomes unavailable or experiences connection issues, it can cause our application to crash or become unresponsive. We want to find a way to avoid this problem while still using Redis OM for our object-oriented interactions.

One potential solution we're considering is to integrate the [Webdis](https://webd.is/) RESTful interface for our on-prem Redis instance with Redis OM. This would allow our application to communicate with the on-prem instance through HTTP requests, making it more resilient to connection issues.

Assuming we can get Webdis to support FT.SEARCH commands, we're wondering if Redis OM can support this use case or if there's an alternative approach we should consider.

slorello89:
👋 @VagyokC4 - this sounds like more or less what's being asked for in #351 (basically a way to full abstract what Redis OM is doing under the hood).

Essentially you'd need to Overload `IRedisConnection`, change it's `Execute`/`ExecuteAsync` to have this alternative where it shoves all the arguments into the path or body of the request, then you would need to be able to marshall the resposnes into a `RedisReply` which evidently is not very initalizable from outside of Redis OM (something we can look at fixing)

VagyokC4:
Woohoo! That sounds like the ticket.  What is your ETA to being able to have the internals abstractable?

VagyokC4:
@slorello89  I'm currently passing around a RedisConnectionProvider and generating my RedisCollections from there.

Now that I want to implement my own RedisConnection, what is the best way to inject it into Redis-OM?

slorello89:
You would need to create your own IRedisConnection and use that to take the commands From Redis OM, you'd initialize your RedisCollection with it.

In this case I think the IRedisConnection would just be a wrapper around and HTTP client where you'd take the arguments from Redis OM and format them into the required payload for the REST endpoint right?

VagyokC4:
> In this case I think the IRedisConnection would just be a wrapper around and HTTP client where you'd take the arguments from Redis OM and format them into the required payload for the REST endpoint right?

Exactly, I'm just going to make an HTTP call in the Execute method.

The problem I'm seeing is the RedisConnectionProvider's connection property doesn't have an easy way to inject the IRedisConnection.
```
        /// <summary>
[...]

PR #357, opened 2023-04-19
Opening up RedisReply constructor

### LINKED ISSUE #351: Missing support for CommandFlags
Hi @slorello89 !

We are using Redis as our main readside datastore. The topology that is currently being used is a master/slave replication configuration with Sentinel. To load-balance the read requests, we need to allow for round-robin access to slave nodes. As per what I found on [this issue](https://github.com/StackExchange/StackExchange.Redis/issues/547) on StackExchange.Redis repo, we need to pass CommandFlag.PreferSlave to our commands to enable the round-robin querying over slaves. By default, the redis client uses the default flag, which I guess is Command.DemandMaster.

I've been looking through RedisOM code, and it seems like it does not support command flags at this stage, what do you suggest? Is there a possibility to add flags support in the near future?

slorello89:
Opening up RedisReply constructor so that you can create a proper extension of IRedisConnection - should fix #351 and #352

### DIFF
--- a/src/Redis.OM/RedisReply.cs
+++ b/src/Redis.OM/RedisReply.cs
@@ -67,7 +67,7 @@ namespace Redis.OM
         /// Initializes a new instance of the <see cref="RedisReply"/> class.
         /// </summary>
         /// <param name="result">the redisResult.</param>
-        internal RedisReply(RedisResult result)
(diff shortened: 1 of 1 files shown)


bsbodden (approved):
LGTM!

Issue #358, opened 2023-04-20
Partial choice of entity is not working without the Id field


AlexanderAnatolich:
Versions:
.Net 6
Redis OM 0.5.0

Redis modules:
Redis search 2.4.16
Redis JSON 2.2.0

**Steps to reproduce:**
Create models, like that:

```
public class NecessarySubEntity {
      public SomeSubModel Field1 { get; set; }
      public SomeSubModel Field2 { get; set; }
}

public class SomeSubModel {
      public long Id { get; set; }
      public string Name { get; set; }
}

[Document(IndexName = "SomeModel", Prefixes = new[] { "SomeModel" }, StorageType = StorageType.Json)]
    public class SomeModel
    {
        [RedisIdField]
        [Indexed(Aggregatable = true, Sortable = true)]
        public long Id { get; set; }

        [Indexed(Aggregatable = true, Sortable = true)]
        public long? Timestamp { get; set; }

        [Indexed(CascadeDepth = 1)]
        public SomeSubModel Field1 { get; set; }
        
        [Indexed(CascadeDepth = 1)]
        public SomeSubModel Field2 { get; set; }
        
        .....
        
        [Indexed]
        public SomeStatus? SomeStatusId { get; set; }
}
```

**Insert 100+ entities to Redis.**

**Execute it the next way:**

```
public async Task<IList<NecessarySubEntity>> GetData(
             int? count,
             Expression<Func<SomeModel, bool>>? predicate = null)
        {
            try
            {
                var source = redisCollection;

                if (predicate != null)
                {
                    source = redisCollection.Where(predicate);
                }

                if (count.HasValue)
                {
                    source = source.Take(count.Value);
                }

                var resulted = source.Select(it => new NecessarySubEntity
                {
                    Field1 = it.Field1 ,
                    Field2 = it.Field2 
                });

               // ToList or ToListAsync doesn't matter
                var result = await resulted.ToListAsync();

                return result;
}
```
**Actual result:**
result is empty

**Expected result:**
result has a list of the values from Redis.

**Comments:**

It start working when I added the Id field to `NecessarySubEntity`, even though I do not select it.
I mean:
```
var resulted = source.Select(it => new NecessarySubEntity
                {
                    Field1 = it.Field1 ,
                    Field2 = it.Field2 
                });
```
OR
```
var resulted = source.Select(it => new NecessarySubEntity
                {
                    Id = it.Id,
                    Field1 = it.Field1 ,
[...]

Issue #361, opened 2023-04-25
Selection into an object with renamed fields is not working


AlexanderAnatolich:
Versions:
.Net 6
Redis OM 0.5.0

Redis modules:
Redis search 2.4.16
Redis JSON 2.2.0

**Steps to reproduce:**
Create models, like that:

```
public class NecessaryEntity {
     public long MyId { get; set; }
     public long? MyTimestamp { get; set; }
}

[Document(IndexName = "SomeModel", Prefixes = new[] { "SomeModel" }, StorageType = StorageType.Json)]
    public class SomeModel
    {
        [RedisIdField]
        [Indexed(Aggregatable = true, Sortable = true)]
        public long Id { get; set; }

        [Indexed(Aggregatable = true, Sortable = true)]
        public long? Timestamp { get; set; }  
        .....
        
        [Indexed]
        public SomeStatus? SomeStatusId { get; set; }
}
```

**Insert any entities to Redis.**

**Execute it the next way:**

```
public async Task<IList<NecessaryEntity>> GetData()
        {
                var resulted = redisCollection.Select(it => new NecessaryEntity
                {
                    MyId = it.Id,
                    MyTimestamp = it.Timestamp 
                });

               // ToList or ToListAsync doesn't matter
                var result = await resulted.ToListAsync();

                return result;
        }
```
**Actual result:**
result is empty

**Expected result:**
The result has a list of the values from Redis.

**Comments:**

Doesn't matter how I tried to select a typed object or anonymized.
I mean:
```
var resulted = redisCollection.Select(it => new NecessaryEntity
                {
                    MyId = it.Id,
                    MyTimestamp = it.Timestamp 
                });
```
OR
```
var resulted = source.Select(it => new
                {
                    MyId = it.Id,
                    MyTimestamp = it.Timestamp 
                });
```

It works only if are fields name matches.
Like:
```
var resulted = source.Select(it => new
                {
                    Id = it.Id,
                    Timestamp = it.Timestamp
                });
```

slorello89:
Hmm It looks like it's pulling the mapped proper names from the wrong side of the expression. Will fix this along with the other Select issue.

PR #363, opened 2023-05-01
Fixing issues with how selects work with Complex types

slorello89:


slorello89:
Should fix #358 and #361

shacharPash (approved):

PR #368, opened 2023-05-02
fixing issue with index serialization when propertyname defined

### LINKED ISSUE #347: [Question] Strange prefix in JSON indexing
https://github.com/redis/redis-om-dotnet/blob/dc2e33e667f0bb458b1c0f4a5432314f153fc2b4/src/Redis.OM/Modeling/RedisSchemaField.cs#L72
Guys, could u please explain me what was a purpose when u were doing such a thing *$"{pathPrefix}{attr.PropertyName}{pathPrefix}"*? I would understand if u use pathPostFix in both places, but when I see '$.key$.' as index identifier I have no idea what to do. Do u suggest to and '$.' as ending to each and every property of my documents I want to index eventually?

slorello89:
Fixes #347 - slight typo in the way the alias' are generate caused this. mostly just tests.


bsbodden (approved):
LGTM!

PR #372, opened 2023-05-04
fixing selects with same names

### LINKED ISSUE #371: Partial choice of entity to exactly object is not working
Parent issue: #358 
Versions:
.Net 6
Redis OM 0.5.1

Redis modules:
Redis search 2.4.16
Redis JSON 2.2.0

**Steps to reproduce:**
Create models, like that:

```
public class NecessarySubEntity {
      public SomeSubModel Field1 { get; set; }
      public SomeSubModel Field2 { get; set; }
}

public class SomeSubModel {
      public long Id { get; set; }
}

[Document(IndexName = "SomeModel", Prefixes = new[] { "SomeModel" }, StorageType = StorageType.Json)]
    public class SomeModel
    {
        [RedisIdField]
        [Indexed(Aggregatable = true, Sortable = true)]
        public long Id { get; set; }

        [Indexed(Aggregatable = true, Sortable = true)]
        public long? Timestamp { get; set; }

        [Indexed(CascadeDepth = 1)]
        public SomeSubModel Field1 { get; set; }
        
        [Indexed(CascadeDepth = 1)]
        public SomeSubModel Field2 { get; set; }
        
        .....
        
        [Indexed]
        public SomeStatus? SomeStatusId { get; set; }
}
```

**Insert any count of entities to Redis.**

**Execute it the next way:**

```
public async Task<IList<NecessarySubEntity>> GetData()
        {
            try
            {
                var resulted = redisCollection.Select(it => new NecessarySubEntity
                {
                    Field1 = it.Field1 ,
                    Field2 = it.Field2 
                });

               // ToList or ToListAsync doesn't matter
                var result = await resulted.ToListAsync();

                return result;
}
```
**Actual result:**
Receive an array with properties that have null values.

**Expected result:**
Receive an array with properties that have necessary values.

**Comments:**

It works when I select the necessary properties to anonymize the object.
I mean:
```
var resulted = source.Select(it => new
                {
                    Field1 = it.Field1 ,
                    Field2 = it.Field2 
                });
```
But when I'm selecting it to exactly object, I receive a null for the properties.
```
[
  {
     Field1 is null,
     Field2 is null
  },
  {
     Field1 is null,
     Field2 is null
  }
]

slorello89:
Fixes #371 - decision to not alias if field name was the same in the object being projected onto as the initial object was apparently ill-advised as it was making it impossible to parse correctly.

### DIFF
--- a/src/Redis.OM/Common/ExpressionTranslator.cs
+++ b/src/Redis.OM/Common/ExpressionTranslator.cs
@@ -627,14 +627,7 @@ namespace Redis.OM.Common
                         if (assignment.Expression is MemberExpression assignmentExpression)
                         {
                             var path = AliasOrPath(t, attr, assignmentExpression);
-                            if (assignmentExpression.Member.Name == binding.Member.Name)
-                            {
-                                returnFields.Add(new (path));
-                            }
-                            else
-                            {
-                                returnFields.Add(new (path, binding.Member.Name));
-                            }
+                            returnFields.Add(new (path, binding.Member.Name));
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


bsbodden (approved):
LGTM!

Issue #116, opened 2022-05-27
Exception "Only a single Generic type is supported on enums for the Hash type" is thrown when projecting arrays


ivan-mankov-u:
I want to create a projection for an array property using Select. When executing a query, exception is thrown with message "Only a single Generic type is supported on enums for the Hash type". Seems it is related to IEnumerable types only, since other types are projected successfully. 

The index created:
`"FT.CREATE" "person-idx" "ON" "Json" "PREFIX" "1" "Person" "SCHEMA" "$.Id" "AS" "Id" "TAG" "SEPARATOR" "|" "$.NickNames[*]" "AS" "NickNames" "TAG" "SEPARATOR" "|"`

Generated search query:
`"FT.SEARCH" "person-idx" "@NickNames:{c}" "LIMIT" "0" "100" "RETURN" "2" "Id" "NickNames"`
   
    [Document(StorageType = StorageType.Json, Prefixes = new[] { nameof(Person) })]
    public class Person
    {
        [RedisIdField]
        [Indexed]
        public string Id { get; set; }
        public string FirstName { get; set; }
        public string LastName { get; set; }
        [Indexed]
        public string[] NickNames { get; set; }
    }

    internal class RedisArrays
    {
        public static async Task Run()
        {
            var provider = new RedisConnectionProvider("redis://localhost:6379");
            provider.Connection.CreateIndex(typeof(Person));

            var collection = provider.RedisCollection<Person>();
            await collection.InsertAsync(new Person { Id = "1", FirstName = "John", LastName = "Smith", NickNames = new[] { "a", "b", "c" } });
            await collection.InsertAsync(new Person { Id = "2", FirstName = "Jane", LastName = "Dowes", NickNames = new[] { "c", "d", "e" } });

            var persons = await collection
                .Where(p => p.NickNames.Contains("c"))
                .Select(p => new { p.Id, p.NickNames })
                .ToListAsync();
        }
    }

zulander1:
it's wired but this code works:
```

customers.Insert(new Customer()
{
    FirstName = "James",
    LastName = "Bond"
});

// Find all customers with the nickname of Jim
var test = customers.Where(x => x.FirstName == "James").Select(x => new test() { FirstName = x.FirstName, LastName = x.LastName }).ToListAsync();
var test2 = customers.Where(x => x.FirstName == "James").Select(x => new { FirstName = x.FirstName, LastName = x.LastName }).ToListAsync();
[...]

PR #402, opened 2023-07-31
Add support for StartsWith/EndsWith, extended support for Contains

zulander1:
fixes issue #116

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,8 @@ We'd love your contributions! If you want to contribute please read our [Contrib
 * [@Jeevananthan](https://github.com/Jeevananthan-23)
 * [@mariusmuntean](https://github.com/mariusmuntean)
 * [@jcreus1](https://github.com/jcreus1)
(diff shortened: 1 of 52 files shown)


zulander1:
This should be break in change since 

- `collection.Where(x => x.Name.Contains("*Ste*")).ToList();` would find word which contains (==) Ste
- `collection.Where(x => x.Name.Contains("Ste")).ToList();` would find word which is an exact match of Ste
- `collection.Where(x => x.Name.Contains("Ste*")).ToList();` would find word which is an startWith of Ste
- `collection.Where(x => x.Name.Contains("*Ste")).ToList();` would find word which is an endWith of Ste

zulander1:
@slorello89 the test are passing but there is a timeout

slorello89 (approved):
Don't think this actually fixes #116 @zulander1 - that's still very much caused by an issue that needs to be addressed in RediSearch, but definitely adds some nice support for StartsWith/EndsWith/Contains.

LGTM 👍

PR #403, opened 2023-08-06
Fix indexing and querying GUID lists

Espleth:
There is a problem when using [Indexed] `List<Guid>` right now - index for them is not created, and querying is not working. This will fix this.
However, as I can see right now, current aprroach is quite bad - you have to manually add new types and it's possible to forget to add it in one place or another. 
And I had to spend a few hours to find out, why `List<Guid>` is not working, there was no indication from RedisOM that I'm doing something wrong.

Also, you probably can't cover all types, that can be indexed: user can use for indexing any signle-line field (DateTime, DateTimeOffset, Enum, Uri, and the list goes on, not to mention it's possible to create custom types). All these types can be serialized as valid JSON single-line fields (string or number)
So, you need to check not the base type, but the JSON-serialized type of the object (with JsonConverter attribute, if it's there). This seems to be a complex task, so I'll leave it to you =)


Espleth:
Also, the Collection type might be, for example, HashSet, or other ICollection, so current approach of detection collections is bad, too.

slorello89 (approved):

Issue #391, opened 2023-06-19
Date literals not converted to Epoch in aggregation set queries.


giabos:
Hello,
I'm getting an Exception while querying DateTime field in aggregation sets.

here is an example to reproduce the issue:

```
using Redis.OM;
using Redis.OM.Modeling;

RedisConnectionProvider provider = new RedisConnectionProvider("redis://localhost:6379");

var now = DateTime.Now;

// OK redis collection query.
var caiColl = provider.RedisCollection<CAI>();
var list = caiColl.Where(a => a.BatchStartDate < now).Take(10).ToList();
// redis query: "FT.SEARCH" "idxCachedAuctionItem" "(@BatchStartDate:[-inf (1687161811514])" "LIMIT" "0" "10"


// KO (same query but in aggregation set).
var cai = provider.AggregationSet<CAI>();
var counts = cai.Where(a => a.RecordShell!.BatchStartDate < now).GroupBy(a => a.RecordShell!.BodyColorGroup).CountGroupMembers().ToArray();

/*

Error:
    System.Exception : Success (not an error)
    Failed on FT.AGGREGATE idxCachedAuctionItem @BatchStartDate:[-inf (19/06/2023 09:32:30] GROUPBY 1 @BodyColorGroup REDUCE COUNT 0 AS COUNT

    ==> Date literal not converted to Epoch

*/


[Document(StorageType = StorageType.Hash, Prefixes = new[] { "CAI" }, IndexName = "idx1")]
public class CAI
{
    [RedisIdField]
    [Indexed]
    public long AuctionId { get; set; }

    [Indexed(Sortable = true)]
    public DateTime BatchStartDate { get; set; }

    [Searchable(Sortable = true)]
    public string BodyColorGroup { get; set; }
}

```

Tested on library version: **0.5.2**.

Thanks for your work and the help!

PR #406, opened 2023-08-11
adding proper translation for DateTimeOffsets

### LINKED ISSUE #395: Incorrect DateTime filtering
Versions:
.Net 6
Redis OM 0.5.2

Redis modules:
Redis search 2.4.16
Redis JSON 2.2.0

**Steps to reproduce:**
**Create models, like that:**

```
[Document(IndexName = "SomeModel", Prefixes = new[] { "SomeModel" }, StorageType = StorageType.Json)]
    public class SomeModel
    {
        [RedisIdField]
        [Indexed(Aggregatable = true, Sortable = true)]
        public long Id { get; set; }

        [Indexed]
        public DateTime EventDate { get; set; }
}
```

**Insert list of entities to Redis**
In Redis you should have the next data:
```
{
Id:1
EventDate: 1687824000000  (6/26/2023 5:00:00 PM)
},
{
Id:2
EventDate: 1687881600000 (6/27/2023 9:00:00 AM)
}
```
**Pass the next filter:**

model.ToDate (6/28/2023 12:00:00 AM) (2023-06-28T00:00:00.000Z) have the next ticks 638235072000000000 
model.FromDate (6/27/2023 12:00:00 AM) (2023-06-27T00:00:00.000Z) have the next ticks 638234208000000000

**Execute it the next way:**
```
public async Task<IList<SomeModel>> GetData(SomeFilterModel model)
        {
            try
            {
                var resulted = redisCollection.Where(it=>it.EventDate <= model.ToDate && it.EventDate >= model.FromDate);

               // ToList or ToListAsync doesn't matter
                var result = await resulted.ToListAsync();

                return result;
}
```
**Actual result:**
Items with id 1 and id 2 are returned

**Expected result:**
The only item with id 2 is returned

**Comments**
The Redis receives an execution command like that:
FT.SEARCH SomeModel ((@EventDate:[-inf 1687910400000]) 
(@EventDate:[1687824000000 inf])) LIMIT 0 1000
I think the problem is related to time convertation. 
Because when I tried to execute the same code from Redis.OM.Modeling.DateTimeJsonConverter, it returns another result:
```
var datetime = new DateTime(638234208000000000)
new DateTimeOffset(datetime).ToUnixTimeMilliseconds() == 1687849200000
```
but in the query, I have another value 1687824000000

slorello89:
closes #395 

As it stands now DateTimeOffsets are not being parsed correctly in queries, DateTimes are converted to UNIX timestamps before being submitted to a query (based on the local timezone). Using `DateTimeOffsets` is the 'more correct' way to run date ranges (as it has all the timezone information needed to make an informed decision. The challenge was that you couldn't actually use those in your queries before. This fixes that.


imansafari1991 (approved):

slorello89:
Determined that this is a pretty similar issue to #391 - so fixing that as well.

shacharPash (approved):

PR #411, opened 2023-08-15
Fix expression constant in contains

### LINKED ISSUE #409: Possibility to Improve TranslateContainsStandardQuerySyntax to include Expression.Constant
I am working with [ChilliCream GraphQL](https://github.com/ChilliCream/graphql-platform/), and I am having issue with "Contains" when the application is calling Contains using `.Call System.Linq.Enumerable.Contains( .Constant(System.String[]), $_s0.FirstName)` since this is Constant, it doesn't fall though [this ](https://github.com/redis/redis-om-dotnet/blob/6399ed5586bfe8ec36fba33e2877d12440088a20/src/Redis.OM/Common/ExpressionParserUtilities.cs#L796)line. you can see the issue with this file:

https://github.com/zulander1/redis_graphql

Query is (https://localhost:7231/graphql/)
```

{
  customer(where: { firstName: { in: ["james"] } }) {
    nodes {
      firstName
      lastName
    }
  }
}

```

zulander1:
Fixes #409

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,8 @@ We'd love your contributions! If you want to contribute please read our [Contrib
 * [@Jeevananthan](https://github.com/Jeevananthan-23)
 * [@mariusmuntean](https://github.com/mariusmuntean)
 * [@jcreus1](https://github.com/jcreus1)
+* [@JuliusMikkela](https://github.com/JuliusMikkela)
+* [@imansafari1991](https://github.com/imansafari1991)
 
 <!-- Logo -->
 [Logo]: images/logo.svg
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ customers.Where(x => x.LastName == "Bond" || x.Age > 65);
 customers.Where(x => x.LastName == "Bond" && x.FirstName == "James");
 
 // Find all customers with the nickname of Jim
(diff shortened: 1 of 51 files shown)


slorello89:
@zulander1 - try with the updates I just made on your branch, think that ought to work. This is a pretty interesting usecase of Redis OM, one thing I noticed though was that it doesn't seem to issue any Select predicates if you don't request the full object - wonder what that's about?

zulander1:
> @zulander1 - try with the updates I just made on your branch, think that ought to work. This is a pretty interesting usecase of Redis OM, one thing I noticed though was that it doesn't seem to issue any Select predicates if you don't request the full object - wonder what that's about?

Yeah, there is an issue with the Select, but not with the predicates, it doesn't want to create a new instance of the object... I didn't have time to look into, will do this week

slorello89:
@zulander1 - I'm assuming it's nothing on Redis OM's end? I just don't see anything in the predicates being pushed down to it.

zulander1:
@slorello89 Wow it's working flawlessly! For the Select predicates we need to UseProjection

vbedekar:
Any timeline when we can see this in a new release

slorello89 (approved):
LGTM 👍

PR #417, opened 2023-11-02
Vector Search and Semantic Caching

slorello89:
Introduces Vector Search and Semantic caching to Redis OM .NET.

One breaking change - replace `string[]` with `object[]` in some key places (e.g. `Execute`/`ExecuteAsync`) as the byte arrays needed for VectorSearch need to be passed in raw. Should be transparent to anyone using the higher-level APIs within Redis OM, but anyone using those raw commands might need to make some adjustments. See README for details as to how to use the new API.


slorello89:
Some broken tests at first (need to obtain keys for the integration tests with OpenAI/HuggingFace/Azure)

slorello89:
FYI @Spartee, @tylerhutcherson, & @banker

bsbodden (approved):
L-very-GTM!

README.md
```diff
@@ -288,6 +288,104 @@ customers.Where(x => x.LastName == "Bond" && x.FirstName == "James");
 customers.Where(x=>x.NickNames.Contains("Jim"));
 ```
 
+### Vectors
+
+Redis OM .NET also supports storing and querying Vectors stored in Redis. 
+
+A `Vector<T>` is a representation of an object that can be transformed into a vector by a Vectorizer.
+
+A `VectorizerAttribute` is the abstract class you use to decorate your Vector fields, it is responsible for defining the logic to convert your Vectors into Embeddings. In the package `Redis.OM.Vectorizers` we provide vectorizers for HuggingFace, OpenAI, and AzureOpenAI to allow you to easily integrate them into your workflows.
```
tylerhutcherson, line 297: the phrase "convert your Vectors into Embeddings" is a bit misleading as those two terms are relatively interchangeable. I think we're essentially talking about the definition of the various vector field attributes like distance metric, data type, dims, etc? Some of those subsumed by the choice of vectorizer for sure

README.md
```diff
@@ -288,6 +288,104 @@ customers.Where(x => x.LastName == "Bond" && x.FirstName == "James");
...
+To define a vector in your model, simply decorate a `Vector<T>` field with and `Indexed` and a `Vectorizer` attribute (in this case we'll use OpenAI):
+
+```cs
+[Document(StorageType = StorageType.Json)]
+public class OpenAIQuery
+{
+    [RedisIdField]
+    public string Id { get; set; }
+
+    [Indexed(DistanceMetric = DistanceMetric.COSINE)]
```
tylerhutcherson, line 310: Can you also show how a few other vector field attributes like index type (HNSW vs FLAT) and related args are set here?

README.md
```diff
@@ -288,6 +288,104 @@ customers.Where(x => x.LastName == "Bond" && x.FirstName == "James");
...
+// hybrid query, pre-filters responses newer than 4 hours, and finds first result within .15
+var ts = DateTimeOffset.Now - TimeSpan.FromHours(4);
+result = collection.First(x=>x.TimeStamp > ts && x.Prompt.VectorRange(queryPrompt, .15));
+```
+
+#### What Happens to the Embeddings?
+
+With Redis OM, the embeddings can be completely transparent to you, they are generated and bound to the `Vector<T>` when you query/insert your vectors. If however you needed your embedding after the insertion/Query, they are available at `Vector<T>.Embedding`, and be queried either as the raw bytes, as an array of doubles or as an array of floats (depending on your vectorizer).
+
+#### Configuration
```
tylerhutcherson, line 366: Add other vector field attribute level configuration details here too?

README.md
```diff
@@ -288,6 +288,104 @@ customers.Where(x => x.LastName == "Bond" && x.FirstName == "James");
...
+| Configuration Parameter            | Description                                   |
+|--------------------------------    |-----------------------------------------------|
+| REDIS_OM_HF_TOKEN                  | HuggingFace Authorization token.              |
+| REDIS_OM_OAI_TOKEN                 | OpenAI Authorization token                    |
+| REDIS_OM_OAI_API_URL               | OpenAI URL                                    |
+| REDIS_OM_AZURE_OAI_TOKEN           | Azure OpenAI api key                          |
+| REDIS_OM_AZURE_OAI_RESOURCE_NAME   | Azure resource name                           |
+| REDIS_OM_AZURE_OAI_DEPLOYMENT_NAME | Azure deployment                              |
+
+### Semantic Caching
```
tylerhutcherson, line 379: LOVE THIS.

Can we show off here how to also set the semantic threshold for the cache check (that's a key component of a semantic cache, as it varies by use case)

src/Redis.OM/Modeling/IndexedAttribute.cs
```diff
@@ -17,7 +17,79 @@ public sealed class IndexedAttribute : SearchFieldAttribute
         /// </summary>
         public bool CaseSensitive { get; set; } = false;
 
+        /// <summary>
+        /// Gets or sets the vector storage algorithm to use. Defaults to Flat (which is brute force).
+        /// </summary>
```
tylerhutcherson, line 22: Awesome. Can we also show index type in the readme?

README.md
```diff
@@ -288,6 +288,104 @@ customers.Where(x => x.LastName == "Bond" && x.FirstName == "James");
...
+    public DateTime TimeStamp { get; set; }
+}
+```
+
+#### Insert Vectors into Redis
+
+With the vector defined in our model, all we need to do is create Vectors of the generic type, and insert them with our model. Using our `RedisCollection`, you can do this by simply using `Insert`:
+
+```cs
+var query = new OpenAIQuery
```
tylerhutcherson, line 329: naming this `query` makes sense given the implied caching use case here. Maybe spell it out a bit so it's clear why we are inserting a "query" object into your vector database?

tylerhutcherson (approved):
Heck yeah. This is awesome Steve. I left a few readme suggestion/ideas, mainly focused on clarity and flow.  Where will this be represented in the docs?

README.md
slorello89: Added details about index definition in the modeling section as that's more or less where it belongs (the configuration section is talking about configuring the vectorizers)

README.md
slorello89: Added below.

README.md
slorello89: Added a couple of other parameters for the index definition, and explained it a bit better in the modeling section.

README.md
slorello89: changed the verbiage a bit - hopefully this is better?

README.md
slorello89: Query is confusing in this context, is `OpenAICompletionResult` & `completionResult` better? (It's something that's not query that you might actually do with these embeddings lol).

Issue #421, opened 2023-11-20
Issue where active document expiration during enumeration causes early return


jrpavoncello:
I think I found an issue with the Redis.OM nuget package where enumerating the collection will return **_early_** if documents are actively expiring while enumerating. 

# Repro Steps:
You can find a somewhat minimum reproducible project here: https://github.com/jrpavoncello/RedisOMTestBed

On latest Redis Stack and Redis.OM, if you:
1. Set up an index (the `POST /Person/CreateIndex` endpoint in the test bed)
![image](https://github.com/redis/redis-om-dotnet/assets/2867338/741ee5ec-ac9a-47b2-a3fc-e7e33158598c)

2. Insert 10,000 documents, let's call each document a  `Person`, and set **half** of them to randomly expire in the next minute (`/Person/CreateIndex` would already have done this for you)
Example:
``` csharp
  // Create an even distribution so that people are expiring in the middle of enumeration
  var expirationMilliseconds = Random.Shared.Next(msToCreateEveryPerson, msToCreateEveryPerson + msToAllowForTesting);
  await database.KeyExpireAsync($"{nameof(Person)}:{person.Id}", TimeSpan.FromMilliseconds(expirationMilliseconds));
```
3. Enumerate the whole index via something like: `redisConnectionProvider.RedisCollection<Person>(2000)` until it's done enumerating (`GET /Person/EnumeratorTest_GetFirstPerson` in the test bed w/ default values)
4. Throw an exception if the number of documents returned is less than 4999 (`10,000 / 2 - 1`)

You will find that the exception gets thrown while documents are actively expiring. The enumerator will stop early with anywhere between 1800-1999 documents (i.e. not 10,000, and less than the chunk size).

I think I've found this is because the [SearchResponse](https://github.com/redis/redis-om-dotnet/blob/fa80a84bb141ab0b6ff44fdf8f2a5e59a8120aee/src/Redis.OM/Searching/SearchResponse.cs#L117) interprets those null documents returned by RedisSearch as (`docArray.Length < 2`) which skips the document so then [RedisCollectionEnumerator](https://github.com/redis/redis-om-dotnet/blob/fa80a84bb141ab0b6ff44fdf8f2a5e59a8120aee/src/Redis.OM/Searching/RedisCollectionEnumerator.cs#L105) interprets "documents count less than the chunk size" as "we're done enumerating, don't get the next chunk". 

[...]

PR #422, opened 2023-11-26
Bugfix - Enumerating collection with documents expiring causes early return

jrpavoncello:
Addresses bug found in #421 i.e. enumerating a collection from a search where the documents were expiring causes an early return resulting in less documents than expected.

@slorello89 Side note - I noticed that the non-generic [SearchResponse](https://github.com/redis/redis-om-dotnet/blob/fa80a84bb141ab0b6ff44fdf8f2a5e59a8120aee/src/Redis.OM/Searching/SearchResponse.cs#L11) would [return an empty Document](https://github.com/redis/redis-om-dotnet/blob/fa80a84bb141ab0b6ff44fdf8f2a5e59a8120aee/src/Redis.OM/Searching/SearchResponse.cs#L25C1-L32C52) i.e. an empty dictionary documentHash, akin to returning null. But the generic `SearchResponse<T>` just skips the document instead. In the same vein, [AggregationResult](https://github.com/redis/redis-om-dotnet/blob/fa80a84bb141ab0b6ff44fdf8f2a5e59a8120aee/src/Redis.OM/Aggregation/AggregationResult.cs#L86) and [AggregationResult\<T\>](https://github.com/redis/redis-om-dotnet/blob/fa80a84bb141ab0b6ff44fdf8f2a5e59a8120aee/src/Redis.OM/Aggregation/AggregationResult.cs#L15) also seem to have a general pattern of returning `null` to their callers (whether internal or external to this package). This seemed like an unexpected functional difference between these classes that all handle a `RedisReply`s so I wanted to call it to your attention for potential future review since these would be out of scope of a patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,7 @@ We'd love your contributions! If you want to contribute please read our [Contrib
 * [@imansafari1991](https://github.com/imansafari1991)
 * [@AndersenGans](https://github.com/AndersenGans)
 * [@mdrakib](https://github.com/mdrakib)
+* [@jrpavoncello](https://github.com/jrpavoncello)
 
 <!-- Logo -->
 [Logo]: images/logo.svg
--- a/src/Redis.OM/Redis.OM.csproj
+++ b/src/Redis.OM/Redis.OM.csproj
@@ -6,9 +6,9 @@
     <RootNamespace>Redis.OM</RootNamespace>
     <Nullable>enable</Nullable>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <PackageVersion>0.5.4</PackageVersion>
-    <Version>0.5.4</Version>
-    <PackageReleaseNotes>https://github.com/redis/redis-om-dotnet/releases/tag/v0.5.4</PackageReleaseNotes>
+    <PackageVersion>0.5.5</PackageVersion>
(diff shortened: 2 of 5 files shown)


jrpavoncello:
@slorello89 Hey Steve, I wasn't sure if you got notifications from PR descriptions so here's an extra ping if not :) I wanted to get your thoughts on the "side note" mentioned there before publishing this PR, I found some stuff that is _probably_ out of scope for this change.

slorello89:
Hi @jrpavoncello - AggregationResults are a bit different in that they will actually return null items based on what's in the pipeline:

```
127.0.0.1:6379> FT.AGGREGATE catalogentry-idx * LIMIT 0 10
 1) (integer) 1
 2) (empty array)
 3) (empty array)
 4) (empty array)
 5) (empty array)
 6) (empty array)
 7) (empty array)
 8) (empty array)
 9) (empty array)
10) (empty array)
11) (empty array)
```

WRT to the non-generic version of `SearchResponse`, I think that I left that there in case someone wanted a more raw version of the Search Response someday, which wasn't a need that ever materialized, so realistically it's probably dead code which should be removed at some point.

slorello89 (approved):
LGTM 👍 thanks for submitting this @jrpavoncello

PR #423, opened 2023-12-04
Bugfix for culture invariant decimal types

### LINKED ISSUE #418: Incorrect behavior when updating decimal values
Hello!

I am using Redis.OM 0.5.4 and I have found that when running on cultures where the decimal separator is not a dot the `Update` and `UpdateAsync` methods don't work correctly on decimal values.

Running the profiler showed that the method is sending the SET command serializing the decimal with a comma instead of a dot. I believe that this should be serialized using a culture-invariant approach to ensure consistent behavior.

Initially, it was difficult to detect this issue because Redis appears to interpret the value as an integer instead of producing an error.

Command captured by the profiler:
> "EVALSHA" "118f0eee57ae15ddfd37bd680e8002358aaf7161" "1" "Person:01HF7QY15KWSKESTE6QD671HAD" "SET" "$.Height" "75,7"

Repro code:
```csharp
using Redis.OM;
using Redis.OM.Modeling;

Thread.CurrentThread.CurrentCulture = new System.Globalization.CultureInfo("es-ES");

await provider.Connection.CreateIndexAsync(typeof(Person));

var collection = provider.RedisCollection<Person>();
var person = new Person() { Name = "Steve", Height = 75.5m };
await collection.InsertAsync(person);

person = collection.First();
person.Height += 0.2m;
await collection.UpdateAsync(person);

[Document(StorageType = StorageType.Json)]
public class Person
{
    [RedisIdField]
    [Indexed]
    [Indexed]
    public decimal Height { get; set; }
}
```

When querying back the object the Height is 75 instead of 75.7.
image

axnetg:
Fixes #418.

This PR solves an issue related to serialization of decimal types (float, double, and decimal) in cultures where the dot is not the decimal number separator, causing updates and some queries to behave inconsistently. I also noticed that aggregations were also affected by this problem, so this fixes that as well.


slorello89:
Hi @axnetg - can you do one more merge with main - seems I don't have permission to push to your branch :). Should fix the build issues.

slorello89 (changes requested):
Hey @axnetg - yeah sorry there was a breaking change in 0.6.0 (basically the core `Execute` method now accepts an array of objects rather than an array of strings, so all these setup bits need to be updated.

PR #424, opened 2023-12-05
Bugfix/vector hash idx creation

slorello89:
Minor fix to index creation for Hash fields (they need to alias `PropertyName.Vector` to `PropertyName` for queries to work correctly)


bsbodden (approved):
LGTM!

PR #435, opened 2024-02-15
correcting unlink return type

### LINKED ISSUE #433: No Status is returned by many commands
Hello, 

Could you please take a look and see if this is legit or if I need to do things differently?

Here is an example of my code: 

        ` string status =  await _redisConnectionProvider.Connection.UnlinkAsync($"student:{studentId}");`

Under the hood in namespace Redis.OM --> public static class RedisCommands

`   
    public static string Unlink(this IRedisConnection connection, string key) => (string) connection.Execute("UNLINK", (object) key);`

Status is always empty string --> ""

OR: I tried to go deeper and I found this result: 

        ` var status= await _redisConnectionProvider.Connection.ExecuteAsync("UNLINK",$"student:{studentId}");`

here status is of type `Task<RedisReply>` and return _internalLong which can be used to check the status but I felt it is not very straightforward

slorello89:
Fixes #433

Unlink should return a long not a string (it does not get cast correctly). technically a breaking change so updating version as well.


bsbodden (approved):
🚀 LGTM!

Issue #441, opened 2024-03-02
Any not support Contains method


gruan01:
Redis.OM version 0.6.1

~~~c#
xxxCollection.Where(h => h.D == date && h.P.Any(pp => bcs.Contains(pp.B))).Take(10).ToListAsync();
~~~
throw exception:
~~~
  消息: 
    System.InvalidCastException : Unable to cast object of type 'System.Linq.Expressions.MethodCallExpression2' to type 'System.Linq.Expressions.BinaryExpression'.

  堆栈跟踪: 
    ExpressionParserUtilities.TranslateAnyForEmbeddedObjects(MethodCallExpression exp, List`1 parameters)
    ExpressionParserUtilities.TranslateMethodStandardQuerySyntax(MethodCallExpression exp, List`1 parameters)
    ExpressionParserUtilities.GetOperandStringForQueryArgs(Expression exp, List`1 parameters, Boolean treatEnumsAsInt, Boolean negate)
    ExpressionTranslator.TranslateBinaryExpression(BinaryExpression binExpression, List`1 parameters)
    ExpressionTranslator.BuildQueryFromExpression(Expression exp, List`1 parameters)
    ExpressionTranslator.TranslateWhereMethod(MethodCallExpression expression, List`1 parameters)
    ExpressionTranslator.BuildQueryFromExpression(Expression expression, Type type, Expression mainBooleanExpression, Type rootType)
    RedisCollectionEnumerator`1.ctor(Expression exp, IRedisConnection connection, Int32 chunkSize, RedisCollectionStateManager stateManager, Expression`1 booleanExpression, Boolean saveState, Type rootType, Type type)
    RedisCollection`1.GetAsyncEnumerator(CancellationToken cancellationToken)
    RedisCollection`1.ToListAsync()
    RedisIndexTest.SearchTest() 行 147
    --- End of stack trace from previous location ---
~~~

if update this method `ExpressionParserUtilities.TranslateAnyForEmbeddedObjects` like the following can resolve this problem:
~~~c#
        private static string TranslateAnyForEmbeddedObjects(MethodCallExpression exp, List<object> parameters)
        {
            var type = exp.Arguments.Last().Type;
            var prefix = GetOperandString(exp.Arguments[0]);
            var lambda = (LambdaExpression)exp.Arguments.Last();

            if (lambda.Body is MethodCallExpression methodCall)
            {
                var tempQuery = TranslateMethodExpressions(methodCall, parameters);
[...]

PR #446, opened 2024-04-26
Bugfix: support method call in Any() on embedded objects array

PrudiusVladislav:
Solves #441, where `Any()` for embedded objects does not support method call expressions (e.g. `Contains()`, `StartsWith()` etc.). So, in addition to `BinaryExpression `translation support, added the handling of `MethodCallExpression `case. So, it would allow such constructions as, for example: `collection.Where(x => x.EmbeddedCollection.Any(a => a.Property.Contains("Value")))`. Previously, only basic BinaryExpressions like the equality check '==' were allowed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -470,6 +470,7 @@ We'd love your contributions! If you want to contribute please read our [Contrib
 * [@mdrakib](https://github.com/mdrakib)
 * [@jrpavoncello](https://github.com/jrpavoncello)
 * [@axnetg](https://github.com/axnetg)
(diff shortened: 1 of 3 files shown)


PrudiusVladislav:
Hi, @slorello89, do you have any updates or feedback regarding the PR?)

slorello89 (approved):
Just added `StartsWith` and `EndsWith` to test case, LGTM 👍

slorello89:
Thanks for adding this @PrudiusVladislav!

PR #450, opened 2024-05-22
Add global query filter functionality

kirollosonsi:
Add ability to have global query filter and disabling it when needed.


kirollosonsi:
@slorello89 What do you think?

slorello89 (commented):
The `IRedisCollection` is intended to be transient, global queries are something that should be handled in your application's DI layer.

kirollosonsi:
yes, Also ef core dbContext registered as scoped by default but still have such methods.

Issue #460, opened 2024-07-12
Thrown exception on every aggregator call


tylerhennessy:
I am running a simple aggregation like this:

var results = provider.AggregationSet<User>().GroupBy(user => user.RecordShell.SessionId).CountGroupMembers().ToList();

It correctly gets the results, but looking in the Debug Output, it is throwing an exception every time.  I debugged into it and it appears like in the Dispose method of the AggregationEnumerator, if the _cursor is not 0, we try and delete the cursor.  The only problem is UseCursor is set to false in this case, so _cursor defaults to -1 and never changes.  Since it is not 0, we try and delete it and it fails with ERR invalid cursor.

Seems like if _cursor is -1 or UseCursor is false in the Dispose method, we should not be trying to delete the cursor and waste the overhead of executing another command and throwing an exception.

tylerhennessy:
I updated the Dispose method to early out if cursor is 0 or -1 and the average latency for the aggregate call went from 40 ms to 10 ms and obviously didn't log exceptions on every call anymore

tylerhennessy:
Thanks for the quick turnaround.  Applied your change locally and verified it fixes my issue.  Looking forward to it getting published.

PR #462, opened 2024-07-15
setting internal cursor to 0 when no cursor is called for

slorello89:


slorello89:
Fixes #460

guyroyse (approved):
Ship it hoss!

PR #471, opened 2024-07-25
Permanent saving of some types

granit1986:
Several types such as `DateTime`, `byte[]` which are different from redis types, were always detected as modified and saved

### DIFF
--- a/test/Redis.OM.Unit.Tests/RediSearchTests/SearchTests.cs
+++ b/test/Redis.OM.Unit.Tests/RediSearchTests/SearchTests.cs
@@ -3863,21 +3863,37 @@ namespace Redis.OM.Unit.Tests.RediSearchTests
         public void SaveDateTimeIfChanged()
         {
             _substitute.ClearSubstitute();
-            _substitute.Execute(Arg.Any<string>(), Arg.Any<object[]>()).Returns(_mockReply);
-
             var dateTime = new DateTime(2024, 01, 01);
+            var timestamp = new DateTimeOffset(dateTime).ToUnixTimeMilliseconds();
+            var mockReply = (RedisReply)new RedisReply[]
+            {
+                new(1),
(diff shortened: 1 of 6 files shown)


granit1986:
@slorello89 I don`t understand what it's happened. Local this tests works fine

slorello89:
@granit1986 - it looks like you are encountering an issue with timezones, I don't think there's anything wrong with thow the code is operating at least insofar as `DateTime` is concerned. A `DateTime` is converted to a `DateTimeOffset` with the local timezone information so it can be converted back and forth to the same `DateTime` as the server that constructed it was using.

I'm able to revert your changes and still have those tests pass so.

granit1986:
@slorello89 

Great. Thank you. I think we can leave it like this.

granit1986:
I was hoping that these edits would make it into the main repository.

slorello89:
@granit1986 - I understand, but your tests pass against the original code base, so it's unclear what's being fixed by them, it sounds more like you were having a timezone issue with your DateTimes then an actual defect in Redis OM.

granit1986:
I understand. I will try fix test for new case

test/Redis.OM.Unit.Tests/RediSearchTests/SearchTests.cs
```diff
@@ -3840,6 +3845,67 @@ public void TestBasicQueryCastToIQueryable()
...
+
+            var dateTime = new DateTime(2024, 01, 01);
+
+            var collection = new RedisCollection<Person>(_substitute);
+            var item = collection.First();
+            item.Date = dateTime;
+            collection.Update(item);
+            
+            var timestamp = new DateTimeOffset(dateTime).ToUnixTimeMilliseconds();
+            _substitute.Received(0).Execute(
```
granit1986, line 3862: @slorello89 i changed test.
here it is implied that the date update should not occur if the value has not changed

PR #472, opened 2024-07-25
feat: adding the ability to update ttl with update

### LINKED ISSUE #430: UpdateAsync() doesnt provide Inputs for the lifetime cache
Are there any plans to have this option available soon so that we can start using this for setting up custom lifetime while calling update async()

slorello89:
Closes #430, adding the ability to set a ttl when updating


guyroyse (approved):
Looks good to me. 🚀🌕

PR #473, opened 2024-07-29
Allowing SearchableAttribute to apply multiple times on one field

slorello89:


guyroyse (approved):
Looks like it always worked, just wasn't a thing you could do. Ship it!

PR #479, opened 2024-09-02
feat: Add the ability to compare existing index definitions to the projected one for a type

abbottdev:
Hi @slorello89,

This PR adds support to be able to have a _VERY_ simple mechanism for the consumer of the library to determine if they need to drop/recreate a given index.

I foresee people using this method like this pseudocode Index drop/create is async I believe, so the consumer would need to do their own polling to wait for the index to properly delete first but:
```csharp
var definition = provider.Connection.GetIndexInfo(typeof(Customer));

if (definition.IndexDefinitionEquals(typeof(Customer)) == false)
{
    provider.Connection.DropIndex(typeof(Customer));
}

provider.Connection.CreateIndex(typeof(Customer));
```

Please forgive me that I've not raised an issue for this beforehand - I'm hoping this PR makes enough sense for us to discuss here.

Rather than making all the internals of `RedisIndex` visible, I've added an extension method that can be used to compare index definitions. 

Happy to discuss implementation/naming/test coverage but figure it's easier to discuss something visual.

**EDIT**: First run of the action failed, looks like `ubuntu-latest` has moved on to use `docker compose` over `docker-compose` (v2) now.

### DIFF
--- a/.github/workflows/dotnet-core.yml
+++ b/.github/workflows/dotnet-core.yml
@@ -14,4 +14,4 @@ jobs:
       - name: fetch-models
         run: sh fetch-models.sh
       - name: execute
-        run: docker-compose -f ./docker/docker-compose.yaml run dotnet
+        run: docker compose -f ./docker/docker-compose.yaml run dotnet
--- a/README.md
+++ b/README.md
@@ -107,18 +107,17 @@ var provider = new RedisConnectionProvider("redis://localhost:6379");
 provider.Connection.CreateIndex(typeof(Customer));
 ```
 
-Redis OM provides limited support for schema and data migration at this time. We provide a small extension method `IndexDefinitionEquals` on the `RedisIndexInfo` type that you may opt in to use to determine when to re-create your indexes when your types change. An example implementation of this would look like:
+Redis OM provides limited support for schema migration at this time. You can check if the index definition in Redis matches your current index definition using the `IsIndexCurrent` method on the `RedisConnection`. Then you may use that output to determine when to re-create your indexes when your types change. An example implementation of this would look like:
 
 ```csharp
 var provider = new RedisConnectionProvider("redis://localhost:6379");
 var definition = provider.Connection.GetIndexInfo(typeof(Customer));
 
-if (definition.IndexDefinitionEquals(typeof(Customer)) == false)
+if (!provider.Connection.IsIndexCurrent(typeof(Customer)))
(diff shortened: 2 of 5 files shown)


abbottdev:
Bump @slorello89. Hope all's OK - not seen any issue activity on this repo for 2 months and this PR has been outstanding for a bit - any chance we can give the Redis .NET repo some love? ❤️ 

cc @benoitdion

obradl:
I support this functionality, it is much needed. I really enjoy Redis OM for .NET so far and this would be a nice improvement imo.

slorello89 (changes requested):
@abbottdev - I think the only thing this really needs is to be able to answer this question directly from the connection. e.g. `IsModelCurrent`

```cs
provider.Connection.IsModelCurrent(typeof(Customer));
```

that way the library just takes care of everything for you without you having to fetch the index definition and then serialize the model yourself.

WDYT?

VagyokC4:
> @abbottdev - I think the only thing this really needs is to be able to answer this question directly from the connection. e.g. `IsModelCurrent`
> 
> ```cs
> provider.Connection.IsModelCurrent(typeof(Customer));
> ```
> 
> that way the library just takes care of everything for you without you having to fetch the index definition and then serialize the model yourself.
> 
> WDYT?

Maybe even a method that keeps the index up to date
```cs
provider.Connection.UpdateIndex(typeof(Customer));
```
using the pattern below:

> ```cs
> var provider = new RedisConnectionProvider("redis://localhost:6379");
> var definition = provider.Connection.GetIndexInfo(typeof(Customer));
> 
> if (definition.IndexDefinitionEquals(typeof(Customer)) == false)
> {
>     provider.Connection.DropIndex(typeof(Customer));
> }
> 
> provider.Connection.CreateIndex(typeof(Customer));
> ```

slorello89 (approved):
@abbottdev - thanks for submitting this! I hope you don't mind, I pushed an update to add a method directly to the connection to check if the indexes match, I think it's good to go now, just awiting CI/CD

VagyokC4:
> @VagyokC4 my big hangup with that approach is that I'm concerned about any implicit index dropping. Perhaps I'm overly cautious, but the drop / create has some implicit downtime (as the index rebuilds), so I'd rather not hide the drop.

I thought I saw somewhere there was an update index which doesn't drop it but merely patches it?

PR #487, opened 2024-10-03
Add ToQueryString method to RedisCollection, RedisAggregationSet

### LINKED ISSUE #486: Provide ToQueryString from RedisCollection and RedisAggregationSet for debugging/logging
It would be useful for debugging and application logging purposes to be able to serialize a complex expression associated with a RedisCollection or RedisAggregationSet without also executing the resulting command.

[For example, EFCore exposes ToQueryString from implementers of IQueryingEnumerable.](https://github.com/dotnet/efcore/blob/23c67d11c4ee1b47de7f31c1c701bd27dc2ab405/src/EFCore/Query/IQueryingEnumerable.cs#L33C12-L33C25)

tgmoore:
Exposes ToQueryString from RedisCollection and RedisAggregationSet.

Resolves #486

The following questions came up while putting this together:

1. Is `ToQueryString` a reasonable name? Is `ToCommandString` more accurate for Redisearch?
2. These methods duplicate behavior found in `AggregationEnumerator.SerializedArgs` and `RedisCollectionEnumerator.RedisCollectionEnumerator`. Is that reasonable, or should there be a common source for that behavior?
3. Should the tests cover less RedisCollection/RedisAggregationSet behavior and just do the least amount necessary to serialize the command to a string?

### DIFF
--- a/src/Redis.OM/Aggregation/RedisAggregationSet.cs
+++ b/src/Redis.OM/Aggregation/RedisAggregationSet.cs
@@ -117,7 +117,7 @@ namespace Redis.OM.Aggregation
             => (await ToListAsync()).ToArray();
 
         /// <summary>
-        /// A string representation of the aggregation command and parameters, a serialization of the Expression.
+        /// A string representation of the aggregation command and parameters, a serialization of the Expression with all parameters explicitly quoted.
         /// Warning: this string may not be suitable for direct execution and is intended only for use in debugging.
         /// </summary>
         /// <returns>A string representing the Expression serialized to an aggregation command and parameters.</returns>
@@ -132,7 +132,9 @@ namespace Redis.OM.Aggregation
                 serializedArgs.Add(_chunkSize.ToString());
             }
 
-            return $"FT.AGGREGATE {string.Join(" ", serializedArgs)}";
+            var quotedArgs = serializedArgs.Select(arg => $"\"{arg}\"");
+
+            return $"\"FT.AGGREGATE\" {string.Join(" ", quotedArgs)}";
         }
 
         private void Initialize(RedisQueryProvider provider, Expression? expression, bool useCursor)
--- a/src/Redis.OM/Aggregation/RedisAggregationSet.cs
+++ b/src/Redis.OM/Aggregation/RedisAggregationSet.cs
@@ -6,6 +6,7 @@ using System.Linq.Expressions;
 using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
+using Redis.OM.Common;
 using Redis.OM.Contracts;
 using Redis.OM.Modeling;
 using Redis.OM.Searching;
@@ -114,6 +115,25 @@ namespace Redis.OM.Aggregation
         /// <returns>a task that will resolve when the array is enumerated.</returns>
         public async ValueTask<AggregationResult<T>[]> ToArrayAsync()
             => (await ToListAsync()).ToArray();
+
+        /// <summary>
+        /// A string representation of the aggregation command and parameters, a serialization of the Expression.
+        /// Warning: this string may not be suitable for direct execution and is intended only for use in debugging.
+        /// </summary>
+        /// <returns>A string representing the Expression serialized to an aggregation command and parameters.</returns>
+        public string ToQueryString()
+        {
+            var serializedArgs = ExpressionTranslator.BuildAggregationFromExpression(Expression, typeof(T)).Serialize().ToList();
+
+            if (_useCursor)
+            {
+                serializedArgs.Add("WITHCURSOR");
+                serializedArgs.Add("COUNT");
+                serializedArgs.Add(_chunkSize.ToString());
+            }
+
+            return $"FT.AGGREGATE {string.Join(" ", serializedArgs)}";
+        }
 
         private void Initialize(RedisQueryProvider provider, Expression? expression, bool useCursor)
         {
--- a/src/Redis.OM/Searching/RedisCollection.cs
+++ b/src/Redis.OM/Searching/RedisCollection.cs
@@ -736,7 +736,10 @@ namespace Redis.OM.Searching
         {
             var query = ExpressionTranslator.BuildQueryFromExpression(Expression, typeof(T), BooleanExpression, RootType);
             query.Limit ??= new SearchLimit { Number = ChunkSize, Offset = 0 };
-            return $"FT.SEARCH {string.Join(" ", query.SerializeQuery())}";
+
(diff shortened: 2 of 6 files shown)


src/Redis.OM/Aggregation/RedisAggregationSet.cs
```diff
@@ -115,6 +116,25 @@ public async ValueTask<List<AggregationResult<T>>> ToListAsync()
...
+            var serializedArgs = ExpressionTranslator.BuildAggregationFromExpression(Expression, typeof(T)).Serialize().ToList();
+
+            if (_useCursor)
+            {
+                serializedArgs.Add("WITHCURSOR");
+                serializedArgs.Add("COUNT");
+                serializedArgs.Add(_chunkSize.ToString());
+            }
+
+            return $"FT.AGGREGATE {string.Join(" ", serializedArgs)}";
```
slorello89, line 135: I think we also need to do a select and wrap all the arguments in double quotes `"` (as well as wrapping the `FT.AGGREGATE` in quotes) 

There's two reasons for this:
1. Syntax / parameterization: the way that Redis protocol works, it requires it's arguments to be parameterized, so all of the serialized arguments above are actually parameterized when they are sent over the wire to Redis (e.g. the protocol tells Redis how many arguments to expect, and then a precise length for each so the query `FT.SEARCH people "@name:{Steve} @city:Satellite" LIMIT 0 100` is actually 6 parameters over the wire: `FT.SEARCH` , `people`, `@name:{Steve} @city:Satellite`, `LIMIT`, `0`, `100`. Particularly in the case of the query string `@name:{Steve} @city:Satellite`, it needs to be made very clear to the RESP serializer (e.g. the redis CLI) that that is the entire parameter, which leads to the ergonomics of it.

2. Ergonomics: I think what's likely to be the most common usage of this feature is to dump a string into a log / console for a user to just copy/paste into the redis-cli / Redis insights to see what it pulls back. Surrounding everything with quotes is how the [MONITOR](https://redis.io/docs/latest/commands/monitor/) works, and this is quite deliberate as it makes it trivial to copy/paste whatever the command was.

test/Redis.OM.Unit.Tests/RediSearchTests/AggregationSetTests.cs
```diff
@@ -659,5 +659,23 @@ public async Task TestNoCursorDelete()
             await _substitute.Received().ExecuteAsync("FT.AGGREGATE", "person-idx", "*", "FILTER", "@TagField == 'foo' && @Address_State == 'FL'");
             await _substitute.DidNotReceive().ExecuteAsync("FT.CURSOR", Arg.Any<object[]>());
         }
+
+        [Fact]
+        public void TestToQueryString()
+        {
+            var collection = new RedisAggregationSet<Person>(_substitute, true, chunkSize: 10000);
+            var command = "FT.AGGREGATE person-idx @Salary:[(30.55 inf] LOAD * APPLY @Address_HouseNumber + 4 AS house_num_modified SORTBY 2 @Age DESC LIMIT 0 10 WITHCURSOR COUNT 10000";
```
slorello89, line 667: See this is what I was worried about above, this command needs quotes to be properly parameterized (from the CLI's perspective)

```
127.0.0.1:6379> FT.AGGREGATE person-idx @Salary:[(30.55 inf] LOAD * APPLY @Address_HouseNumber + 4 AS house_num_modified SORTBY 2 @Age DESC LIMIT 0 10 WITHCURSOR COUNT 10000
(error) Unknown argument `inf]` at position 1 for <main>
```

vs (mind you I've not added any data):

```
127.0.0.1:6379> FT.AGGREGATE person-idx "@Salary:[(30.55 inf]" LOAD * APPLY "@Address_HouseNumber + 4" AS house_num_modified SORTBY 2 @Age DESC LIMIT 0 10 WITHCURSOR COUNT 10000
1) 1) (integer) 0
2) (integer) 0
```

IMO the expected output should be:

```
"FT.AGGREGATE" "person-idx" "@Salary:[(30.55 inf]" "LOAD" "*" "APPLY" "@Address_HouseNumber + 4" "AS" "house_num_modified" "SORTBY" "2" "@Age" "DESC" "LIMIT" "0" "10" "WITHCURSOR" "COUNT" "10000"
```

src/Redis.OM/Searching/RedisCollection.cs
```diff
@@ -731,6 +731,14 @@ public IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToke
             return new RedisCollectionEnumerator<T>(Expression, provider.Connection, ChunkSize, StateManager, BooleanExpression, SaveState, RootType, typeof(T));
         }
 
+        /// <inheritdoc/>
+        public string ToQueryString()
+        {
+            var query = ExpressionTranslator.BuildQueryFromExpression(Expression, typeof(T), BooleanExpression, RootType);
+            query.Limit ??= new SearchLimit { Number = ChunkSize, Offset = 0 };
+            return $"FT.SEARCH {string.Join(" ", query.SerializeQuery())}";
```
slorello89, line 739: same comment regarding quotations surrounding the arguments here.

slorello89 (changes requested):
Hi @tgmoore, first thanks for opening this! I think we just need to wrap all the arguments and the command names in double quotes - and this should be pretty much good to go. To your questions:

1. `ToCommandString` is probably the more correct name insofar as defining what it does, however given that EF Core uses `ToQueryString`, I think it's very probably better to leave it as `ToQueryString`
2.  This is a perfectly reasonable level of duplication 
3. You probably don't need to do all the myriad set of different things, but it's not wrong to do it.

src/Redis.OM/Aggregation/RedisAggregationSet.cs
tgmoore: Will do. Thank you for your clear answers to my questions and additional comprehensive feedback! I really appreciate your time and attention.

test/Redis.OM.Unit.Tests/RediSearchTests/AggregationSetTests.cs
tgmoore: Makes sense. Thanks for the link to MONITOR. This feedback helped me to understand some of the conversations I've seen on the Discord and in other issues about serialization and escaping special characters.

src/Redis.OM/Searching/RedisCollection.cs
tgmoore: I've made this change as well. Thank you!

slorello89 (approved):
@tgmoore LGTM 👍  - thank you for submitting this!

PR #493, opened 2024-10-10
fixing issue with multiple where predicates paired with a select

### LINKED ISSUE #489: Incorrect filtering
Hello.
I found a bug.
When we use complex filtering with multiple Where and then call Select ExpressionTranslator translated only first Where.
Following code

```
[Fact]
public void TestSelects()
{
    var obj = new Person { Name = "steve", Age = 33 };
    var collection = new RedisCollection<Person>(_connection);
    collection.Insert(obj);
    var query = collection.Where(x => x.Age == obj.Age);
    var res = query.Where(x => x.Name == "steve").Select(x => x.Age).ToList();
    Assert.NotEmpty(res);
    collection.Delete(obj);
}
```

will translate to 
`"FT.SEARCH" "person-idx" "(@Age:[33 33])" "LIMIT" "0" "100" "RETURN" "1" "Age"`
instead
`"FT.SEARCH" "person-idx" "((@Age:[33 33]) (@Name:"steve"))" "LIMIT" "0" "100"`

slorello89:
Fixes #489 - adding selects interferers with the multi-filter building portion of the query parser, so needed to find another way to mange it here.

### DIFF
--- a/src/Redis.OM/Common/ExpressionTranslator.cs
+++ b/src/Redis.OM/Common/ExpressionTranslator.cs
@@ -229,7 +229,7 @@ namespace Redis.OM.Common
                                 query.GeoFilter = ExpressionParserUtilities.TranslateGeoFilter(exp);
                                 break;
                             case "Where":
-                                query.QueryText = TranslateWhereMethod(exp, parameters);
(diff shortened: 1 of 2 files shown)


guyroyse (approved):
Also LGTM.

PR #496, opened 2024-10-21
Bugfix/comprehensive index equality

### LINKED ISSUE #494: IsIndexCurrentAsync always returns false
Hi, 

It seems like `IsIndexCurrentAsync ` always returns false when I try the new feature. In this example the code hits the lines with DropIndexAsync and CreateIndexAsync for every run.  I first run the sample and the index is created. Then I do another run and IsIndexCurrentAsync return false when there is no changes in the model (an the index exists from previous run). 

```
using Redis.OM;
using Redis.OM.Modeling;

var isIndexCurrentAsync = await provider.Connection.IsIndexCurrentAsync(typeof(Customer));

if (!await provider.Connection.IsIndexCurrentAsync(typeof(Customer)))
{
    await provider.Connection.DropIndexAsync(typeof(Customer));
    await provider.Connection.CreateIndexAsync(typeof(Customer));
}


[Document(StorageType = StorageType.Json,  Prefixes = new[] { "test" })]
public class Customer
{
    [Indexed] public string FirstName { get; set; }
    [Indexed] public string LastName { get; set; }
    public string Email { get; set; }
    [Indexed(Sortable = true)] public int Age { get; set; }
    [Indexed] public string[] NickNames { get; set; }
}
```

Versions:
Redis OM v. 0.7.5
RediSearch v. 2.8.13
RedisJson v. 2.6.10

slorello89:
Closes #494 

#479 added support for checking the equality of the current state of your model in your code versus what is in Redis, however there are a number of cases it did not account for which I missed during my review (e.g Vector fields, the special arguments from text fields, special arguments in the Document Attribute) one of which someone tripped over in #494. This PR should address all of these.

There is still the special case where it will not be able to definitively tell whether or not a index definition matches as several fields from the index definition are not returned from FT.INFO, namely `STOPWORDS`, `PHONETIC`, `EF_RUNTIME`, `EPSILON`,  in which ase Redis OM will simply throw an `ArgumentException`.

### DIFF
--- a/src/Redis.OM/Modeling/RedisIndex.cs
+++ b/src/Redis.OM/Modeling/RedisIndex.cs
@@ -20,27 +20,86 @@ namespace Redis.OM.Modeling
         public static bool IndexDefinitionEquals(this RedisIndexInfo redisIndexInfo, Type type)
         {
             var serialisedDefinition = SerializeIndex(type);
-            var existingSet = redisIndexInfo.Attributes?.Select(a => (Property: a.Attribute!, a.Type!)).OrderBy(a => a.Property);
             var isJson = redisIndexInfo.IndexDefinition?.Identifier == "JSON";
 
+            var currentOffset = 0;
             if (serialisedDefinition.Length < 5)
             {
                 throw new ArgumentException($"Could not parse the index definition for type: {type.Name}.");
             }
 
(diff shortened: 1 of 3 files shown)


bsbodden (approved):
LGTM!

PR #497, opened 2024-10-21
fixing issue with timespans having fractional millisecond values

slorello89:
Fixes issues with timespans having fractional milliseconds values. Came up in a response to #430. Timespans have a TotalMilliseconds which is a double and was not properly rounded when it was fractional.


bsbodden (approved):
LGTM, had to the same in Spring

PR #506, opened 2024-11-26
Queryable null and Empty Strings

slorello89:
With the GA of RediSearch 2.10, we are now able to support querying [Null and Empty Strings](https://redis.io/docs/latest/develop/interact/search-and-query/indexing/#index-missing-or-empty-values).

This a breaking change, as the default configuration (which enables searchable null and empty strings), Requires RediSearch 2.10, You can disable this behavior by setting `IndexEmptyAndMissing` in the `SearchFieldAttribute` (either `Indexed` or `Searchable` attributes which will allow this to work with RediSearch < 2.10.


bsbodden (approved):
Nice!

PR #519, opened 2025-02-07
Fix edge-case boolean deserialization in RedisObjectHandler.SendToJson

### LINKED ISSUE #518: JsonException when deserializing projection from FT.SEARCH of JSON value
While testing performance of a local project using RedisOM and GraphQL, I encountered a difference in the behavior of Redis that leads to a System.Text.Json.JsonException during deserialization. Specifically, when enumerating the RedisCollection<T> without projection, I encountered no issues; an example of the Redis query and response is below:
```
Query: FT.SEARCH graph3_user_index * LIMIT 0 10 SORTBY UserId ASC
Response (first record):
{
  "UserId": "<obfuscated>",
  "PrincipalName": "<obfuscated>",
  "Email": "nsqsltsac@laste.ml",
  "EmailType": "signInName",
  "FirstName": "nsqsltsac",
  "LastName": "laste",
  "DisplayName": "nsqsltsac laste",
  "Credentials": "none",
  "Phone": "1234567890",
  "Fax": "1234567890",
  "IsApproved": false,
  "IsExternal": false,
  "CreationType": "LocalAccount",
  "CreatedOn": 1650361275000
}
```
Running a similar query with a projection from GraphQL against the same Redis Db produced a different result, however, as well as the exception:
```
Query: FT.SEARCH graph3_user_index * LIMIT 0 11 RETURN 18 UserId AS UserId Email AS Email FirstName AS FirstName LastName AS LastName DisplayName AS DisplayName IsApproved AS IsApproved SORTBY UserId ASC
Response (first record):
[
  "UserId",
  "<obfuscated>",
  "Email",
  "nsqsltsac@laste.ml",
  "FirstName",
  "nsqsltsac",
  "LastName",
  "laste",
  "DisplayName",
  "nsqsltsac laste",
  "IsApproved",
  "0"
]
```
Of particular not is that in this case, IsApproved was returned as a number instead of as a string. During deserialization, this resulted in the JsonException, with the inner exception "InvalidOperationException: Cannot get the value of a token type 'Number' as a boolean".

Is this a known issue? I've created a fork of RedisOM and added some code to https://github.com/redis/redis-om-dotnet/blob/7f43b7bcedb833ea5386db2d1e634cf889181ca7/src/Redis.OM/RedisObjectHandler.cs#L500 to handle the possibility of both string and numerical representations of a boolean value, but I'm not sure if I should be submitting it.
```diff
-                    ret += $"\"{propertyName}\":{((string)hash[lookupPropertyName]).ToLower()},";
+                    var propValue = ((string)hash[lookupPropertyName]).ToLower();
+                    propValue = propValue == "0"
+                        ? "false"
+                        : propValue == "1"
+                            ? "true"
+                            : propValue;
+                    ret += $"\"{propertyName}\":{propValue},";
```

mfaulcon:
Resolves #518, addressing a JSON deserialization issue relating to a discrepancy in the return value from FT.SEARCH when the result is projected; this issue was affecting GraphQL and OData queries, although was not limited to these use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -500,6 +500,7 @@ We'd love your contributions! If you want to contribute please read our [Contrib
 * [@ahmedisam99](https://github.com/ahmedisam99)
 * [@kirollosonsi](https://github.com/kirollosonsi)
 * [@tgmoore](https://github.com/tgmoore)
+* [@mfaulcon](https://github.com/mfaulcon)
 
 <!-- Logo -->
 [Logo]: images/logo.svg
--- a/src/Redis.OM/RedisObjectHandler.cs
+++ b/src/Redis.OM/RedisObjectHandler.cs
@@ -497,7 +497,13 @@ namespace Redis.OM
                         continue;
                     }
 
-                    ret += $"\"{propertyName}\":{((string)hash[lookupPropertyName]).ToLower()},";
+                    var propValue = ((string)hash[lookupPropertyName]).ToLower();
+                    propValue = propValue == "0"
+                        ? "false"
+                        : propValue == "1"
+                            ? "true"
+                            : propValue;
+                    ret += $"\"{propertyName}\":{propValue},";
                 }
                 else if (type.IsPrimitive || type == typeof(decimal) || type.IsEnum)
                 {


slorello89:
Thank you for submitting this @mfaulcon!

Issue #530, opened 2025-05-17
Not effect Condition in function First/FirstOrDefault of LINQ


SMAH1:
Redis.OM: Version 1.0.0
Linq2db: 5.4.1

**Result of**:
```
string name = "John";
IQueryable query = _provider.RedisCollection<Data>(false);
var res  = await query.FindAsync(x => x.Name == name )
```

**Deferent of**:
```
string name = "John";
IQueryable query = _provider.RedisCollection<Data>(false);
var res  = await query.Where(x => x.Name == name ).FindAsync()
```

**Reason**:
Condition of `x => x.Name == name` is ignored when in `FirstAsync/FirstOrDefaultAsync`.

**Note1**:
My Entity:
```
[Document(StorageType = StorageType.Json, Prefixes = new[] { "W:Data:DB" }, IndexName = "data-idx")]
public class Data : _UuidWithTimestampSupport,
{
    public WcaAdminUser()
    {
        Name = string.Empty;
        Nickname = string.Empty;
        Password = string.Empty;
        ...
    }

    /// <summary>
    /// Name of User
    /// </summary>
    [StringLength(64)]
    [Indexed()]
    public string Name { get; set; }

    /// <summary>
    /// Nickname of User
    /// </summary>
    [StringLength(64)]
    [Indexed]
    public string Nickname { get; set; }

   ...
```
**Note2**:
I cloned Redis.OM project and run test and check `FirstAsync/FirstOrDefaultAsync`. All is OK.
I try build this case but I can not yet!

SMAH1:
I think if Redis.OM does not support this situation, the programmer will be notified by Throw Exception.

PR #531, opened 2025-05-17
Linq first function support condition

SMAH1:
See #530


SMAH1:
Something is mistake!
Close and retry again.

Issue #536, opened 2025-06-10
Exception while executing search


mfaulcon-mynexus:
I have a CachedUser class (defined below) for which I'm trying to execute a string.contains query on one of the member fields, during which I'm getting an exception. I've listed the code and runtime values below. If I perform straight equality searches, it appears to work correctly.

**CachedUser.cs**
```
[Document(StorageType = StorageType.Json, Prefixes = [Prefix], IndexName = Prefix + ":idx")]
public class CachedUser : IEntity<Guid>
{
    public const string Prefix = "graph3:user";

    [Indexed(Sortable = true, IndexEmptyAndMissing = false)]
    [RedisIdField]
    public Guid UserId { get; set; }

    [Indexed(IndexEmptyAndMissing = false)]
    public string? Email { get; set; }

    [Indexed(Sortable = true, IndexEmptyAndMissing = false)]
    public string? FirstName { get; set; }

    [Searchable(Sortable = true, IndexEmptyAndMissing = false)]
    public string? LastName { get; set; }

    [Indexed(Sortable = true, IndexEmptyAndMissing = false)]
    public string? DisplayName { get; set; }

    [Indexed(IndexEmptyAndMissing = false)]
    public string? Credentials { get; set; }

    // Irrelevant properties removed

    [JsonIgnore]
    public Guid Id => UserId;
}
```

The Index being created (formatted for readability):
```
FT.CREATE graph3:user:idx ON Json PREFIX 1 graph3:user SCHEMA 
  $.UserId AS UserId TAG SEPARATOR | SORTABLE
  $.Email AS Email TAG SEPARATOR |
  $.FirstName AS FirstName TAG SEPARATOR | SORTABLE
  $.LastName AS LastName TEXT SORTABLE
  $.DisplayName AS DisplayName TAG SEPARATOR | SORTABLE
  $.Credentials AS Credentials TAG SEPARATOR |
```

The expression being executed at runtime:
```
.Where(u => u.contains("Fal"))
.Select(cu => new CachedUser()
{
  UserId = cu.UserId,
  Email = cu.Email,
  FirstName = cu.FirstName,
  LastName = cu.LastName,
  DisplayName = cu.DisplayName
})
.Take(11);
```

The Redis search the expression is being translated to:
```
[...]

Issue #514, opened 2025-01-28
Error on filtering by indexed nullable DateTime


ALDamico:
I'm storing JSON documents in my Redis instance where one of the fields is a Nullable<DateTime> marked with an Indexable attribute.

Inserting and updating the model works flawlessly, however, if I attempt to filter the IRedisCollection where all these properties are null, I get a NullReferenceException:

```System.NullReferenceException: Object reference not set to an instance of an object.
   at Redis.OM.Common.ExpressionParserUtilities.ValueToString(Object value)
   at Redis.OM.Common.ExpressionParserUtilities.GetOperandStringForQueryArgs(Expression exp, List`1 parameters, Boolean treatEnumsAsInt, Boolean negate, Boolean treatBooleanMemberAsUnary)
   at Redis.OM.Common.ExpressionTranslator.TranslateBinaryExpression(BinaryExpression binExpression, List`1 parameters)
   at Redis.OM.Common.ExpressionTranslator.BuildQueryFromExpression(Expression exp, List`1 parameters)
   at Redis.OM.Common.ExpressionTranslator.TranslateWhereMethod(MethodCallExpression expression, List`1 parameters)
   at Redis.OM.Common.ExpressionTranslator.BuildQueryFromExpression(Expression expression, Type type, Expression mainBooleanExpression, Type rootType)
   at Redis.OM.Searching.RedisCollectionEnumerator`1..ctor(Expression exp, IRedisConnection connection, Int32 chunkSize, RedisCollectionStateManager stateManager, Expression`1 booleanExpression, Boolean saveState, Type rootType, Type type)
   at Redis.OM.Searching.RedisCollection`1.GetAsyncEnumerator(CancellationToken cancellationToken)
   at Redis.OM.Searching.RedisCollection`1.ToListAsync()
   at RedisOmNestedObject.Program.Main(String[] args) in C:\Users\Luciano Damico\RiderProjects\RedisOmNestedObject\RedisOmNestedObject\Program.cs:line 31
   at RedisOmNestedObject.Program.<Main>(String[] args)
```

From what I've been able to grok, it looks like the library doesn't create these keys if the value is null, but when the .Where() LINQ method is called, it takes their existence for granted and the ValueToString method throws an exception because it calls the GetType method on a null value.

I'm not sure this is a bug, or I'm simply using Redis wrong (may well be!).

As a workaround, materializing the query to a list and then filtering by null values works, but it feels like a very hacky solution.

PR #538, opened 2025-06-18
Resolve LINQ expression translation for ==/!= null

mfaulcon:
Fixes an issue with LINQ BinaryExpression translation where negation of the search term was not being correctly applied when the right term resolves to null (#536, #514)

Addresses a NullReferenceException when translating an expression involving a literal null value (#536, #514)


mfaulcon:
@slorello89 - Anything else I need to do for this PR? Having this go through would help me immensely.

PR #552, opened 2026-01-27
Fix `INDEXMISSING` on non-tag fields

FICTURE7:
`INDEXMISSING` should be supported on all Redis field types, but Redis.OM do not handle this case when creating indexes.

> INDEXMISSING for **all field types**, introduced in v2.10, allows you to search for missing values, that is, documents that do not contain a specific field. Note the difference between a field with an empty value and a document with a missing value. By default, missing values are not indexed. [1](https://redis.io/docs/latest/commands/ft.create/#required-arguments) 

For example in the following:

```csharp
using Redis.OM;
using Redis.OM.Modeling;

connection.CreateIndex(typeof(Person));

var now = DateTime.UtcNow;
var collection = provider.RedisCollection<Person>();
var persons = await collection
    .Where(p => p.UpdatedAt != null && p.UpdatedAt < now)
    .ToListAsync();

[Document(StorageType = StorageType.Json)]
public class Person
{
    [Indexed]
    [RedisIdField]
    public string? Id { get; set; }

    [Indexed]
    public DateTime? UpdatedAt { get; set; }
}
```

Previously, `CreateIndex` would not properly mark the `UpdatedAt` field with `INDEXMISSING` when creating the index. Which would then cause the `ismissing(@updatedat)` clause in the query to fail:

```
'ismissing' requires defining the field with 'INDEXMISSING' at offset 13 near UpdatedAt
Failed on FT.SEARCH person-idx (-(ismissing(@UpdatedAt)) (@UpdatedAt:[-inf (1769513519098])) DIALECT 2 LIMIT 0 100
```

This patch makes fields marked with `IndexEmptyAndMissing` be marked with `INDEXMISSING` on index creation. If the mapped Redis type is `NUMERIC`, `INDEXMISSING` will only be added if its the .NET type is nullable. The rationale is that only nullable types can be missing, i.e `int Age { get; set; }` cannot be missing, while `int? Age { get; set; }` be missing (in the standard .NET type system at least).

---

> [!NOTE]
> **Medium Risk**
> Changes default index schemas for indexed nullable numerics, geo, and vectors; existing indexes may need recreation for ismissing queries to work.
> 
> **Overview**
> Fixes **FT.CREATE** schema generation so **`INDEXMISSING`** is emitted for **TEXT**, **TAG**, **VECTOR**, **NUMERIC**, and **GEO** fields when **`IndexEmptyAndMissing`** is enabled (the default on **`[Indexed]`**), matching Redis 2.10+ behavior for **`ismissing()`** queries.
> 
> **`CommonSerialization`** now receives the full property type so nullable **`int?`**, **`DateTime?`**, **`GeoLoc?`**, etc. get **`INDEXMISSING`** on **NUMERIC**/**GEO** mappings; **VECTOR** fields get **`INDEXMISSING`** after vector args. **`RedisIndex.IndexDefinitionEquals`** is updated to treat **`INDEXMISSING`** on **VECTOR**, **NUMERIC**, and **GEO** when comparing live index metadata.
> 
> Functional and serialization tests cover nullable numeric/date queries (JSON and Hash) and updated expected **`FT.CREATE`** argument lists.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d8e799e5024639119c0c2a500ab116180c1852a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


slorello89 (approved):
LGTM 👍 Thanks you @FICTURE7 !

